PR #4, opened 2025-11-23
Add shared drives support to all Google Drive API operations

sioquim:
This commit adds support for Google Shared Drives (formerly Team Drives) by including the required API parameters across all file operations.

Changes:
- Added includeItemsFromAllDrives: true to all files.list() calls (5 instances)
- Added supportsAllDrives: true to all Drive API operations (23 instances)
- Updated operations: list, create, get, update, export, and delete

This fixes the issue where the MCP server could only access 'My Drive' files and would fail or return empty results when trying to access Shared Drive content.

Operations now support:
- Listing files in shared drives
- Searching across shared drives
- Creating/updating files in shared drives
- Moving/renaming files in shared drives
- Reading content from shared drive files

Resolves inability to access shared drive folders as documented in Google Drive API: https://developers.google.com/workspace/drive/api/reference/rest/v3/files/list#parameters

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -260,7 +260,9 @@ async function checkFileExists(name: string, parentFolderId: string = 'root'): P
 // INPUT VALIDATION SCHEMAS
 // -----------------------------------------------------------------------------
 const SearchSchema = z.object({
-  query: z.string().min(1, "Search query is required")
+  query: z.string().min(1, "Search query is required"),
+  pageSize: z.number().int().min(1).max(100).optional(),
+  pageToken: z.string().optional()
 });
 
 const CreateTextFileSchema = z.object({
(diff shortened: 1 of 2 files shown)


sioquim:
@piotr-agier  let me know if there are any issues with this PR

src/index.ts
```diff
@@ -1864,7 +1887,8 @@ server.setRequestHandler(CallToolRequestSchema, async (request) => {
         await drive.files.update({
           fileId: spreadsheet.data.spreadsheetId || '',
           addParents: parentFolderId,
-          fields: 'id, name, webViewLink'
```
piotr-agier, line 1877: Not sure if this would cause any issues but I'd add `removeParents: 'root',` here to prevent multi-parenting.

piotr-agier (commented):
Update README.md mentioning shared drives support.

piotr-agier (approved):

PR #12, opened 2026-01-17
feat: add 28 new tools for Calendar, Docs, Sheets, Comments, Drive

ItsiW:
## Summary

This PR adds 28 new tools to enhance the MCP's capabilities:

### Calendar (6 tools)
- `listCalendars` - List all accessible calendars
- `getCalendarEvents` - Get events with filtering
- `getCalendarEvent` - Get a single event by ID
- `createCalendarEvent` - Create events with Google Meet support
- `updateCalendarEvent` - Update existing events
- `deleteCalendarEvent` - Delete events

### Google Docs Editing (5 tools)
- `insertTable` - Insert tables at specified index
- `editTableCell` - Edit table cell content and style
- `insertImageFromUrl` - Insert images from URLs
- `insertLocalImage` - Upload and insert local images
- `applyParagraphStyle` - Apply paragraph formatting (headings, alignment, spacing)

### Comments (5 tools)
- `listComments` - List all comments in a document
- `getComment` - Get comment with full reply thread
- `addComment` - Add comment anchored to text range
- `replyToComment` - Reply to existing comments
- `deleteComment` - Delete comments

### Drive Discovery (2 tools)
- `listGoogleDocs` - List documents with optional filtering
- `getDocumentInfo` - Get detailed document metadata

### Formatting (3 tools)
- `applyTextStyle` - Apply bold, italic, colors, fonts, links to text ranges
- `formatGoogleDocText` - Text formatting with RGB colors
- `formatGoogleDocParagraph` - Paragraph formatting with alignment, spacing

### Sheets Management (4 tools)
- `getSpreadsheetInfo` - Get metadata and list all sheets/tabs
- `appendSpreadsheetRows` - Append rows to end of a sheet
- `addSpreadsheetSheet` - Add new tab to spreadsheet
- `listGoogleSheets` - List spreadsheets in Drive with filtering

### Drive Operations (1 tool)
- `copyFile` - Copy any Drive file with custom name and location

## Changes
- Added calendar service initialization with proper OAuth scopes
- Added 28 tool schemas using Zod validation
- Added tool implementations with proper error handling
- Updated auth/server.ts with calendar.events and calendar.readonly scopes

## Test plan
- [x] Built successfully with `npm run build`
- [x] Tested Calendar tools (create, update, delete events)
- [x] Tested Docs editing tools (tables, images, formatting)
- [x] Tested Comments tools (add, reply, delete)
- [x] Tested Discovery tools (list, info)
- [x] Tested Sheets tools (getSpreadsheetInfo, appendRows, addSheet, listSheets)
- [x] Tested copyFile

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -118,8 +118,9 @@ async function resolvePath(pathStr: string): Promise<string> {
 
   for (const part of parts) {
     if (!part) continue;
+    const escapedPart = part.replace(/\\/g, "\\\\").replace(/'/g, "\\'");
     let response = await drive.files.list({
-      q: `'${currentFolderId}' in parents and name = '${part}' and mimeType = '${FOLDER_MIME_TYPE}' and trashed = false`,
+      q: `'${currentFolderId}' in parents and name = '${escapedPart}' and mimeType = '${FOLDER_MIME_TYPE}' and trashed = false`,
       fields: 'files(id)',
       spaces: 'drive',
       includeItemsFromAllDrives: true,
@@ -631,7 +632,8 @@ async function uploadImageToDriveHelper(
   const uploadResponse = await drive.files.create({
     requestBody: fileMetadata,
     media: media,
-    fields: 'id,webViewLink,webContentLink'
+    fields: 'id,webViewLink,webContentLink',
+    supportsAllDrives: true
   });
 
   const fileId = uploadResponse.data.id;
@@ -651,7 +653,8 @@ async function uploadImageToDriveHelper(
   // Get the webContentLink
   const fileInfo = await drive.files.get({
     fileId: fileId,
-    fields: 'webContentLink'
+    fields: 'webContentLink',
+    supportsAllDrives: true
   });
 
   const webContentLink = fileInfo.data.webContentLink;
@@ -5698,7 +5701,8 @@ server.setRequestHandler(CallToolRequestSchema, async (request) => {
         if (args.uploadToSameFolder !== false) {
           const fileInfo = await drive.files.get({
             fileId: args.documentId,
-            fields: 'parents'
+            fields: 'parents',
+            supportsAllDrives: true
           });
           parentFolderId = fileInfo.data.parents?.[0];
         }


piotr-agier:
Impressive scope. It overlaps with PRs #16/#17/#18/#20. If we take this, we should still review for surface‑area risks (auth scopes, tool list size) and maybe split into smaller PRs for safer review/merge. Let me know if you’re open to splitting or we can cherry‑pick subsets.

ItsiW:
Rebased!

src/index.ts
```diff
@@ -889,2768 +1662,4425 @@ server.setRequestHandler(ListToolsRequestSchema, async () => {
...
-        // Move to trash instead of permanent deletion
-        await drive.files.update({
-          fileId: args.itemId,
-          requestBody: {
-            trashed: true
-          },
-          supportsAllDrives: true
+        let queryString = "mimeType='application/vnd.google-apps.spreadsheet' and trashed=false";
+        if (args.query) {
+          queryString += ` and (name contains '${args.query}' or fullText contains '${args.query}')`;
```
piotr-agier: The `args.query` value is interpolated directly into the Drive API query string without escaping. A value containing a single quote (') breaks out of the string literal and lets the caller inject arbitrary query predicates, e.g. bypassing the mimeType filter or surfacing trashed files.

The existing search handler at line 2748 already does this correctly:
`const escapedQuery = userQuery.replace(/\\/g, "\\\\").replace(/'/g, "\\'");`

Could we apply the same escaping here and in `listGoogleDocs` at line 6006?

src/index.ts
```diff
@@ -274,6 +286,527 @@ async function checkFileExists(name: string, parentFolderId: string = 'root'): P
...
+    media: media,
+    fields: 'id,webViewLink,webContentLink'
+  });
+
+  const fileId = uploadResponse.data.id;
+  if (!fileId) {
+    throw new Error('Failed to upload image to Drive - no file ID returned');
+  }
+
+  // Make the file publicly readable
```
piotr-agier, line 644: This sets role: 'reader', type: 'anyone' on every uploaded image, making it publicly accessible to anyone with the URL. Since insertLocalImage is the only caller, users inserting a private image into a private
doc may not expect it to become public.

A couple of options:
1. Use the document's existing sharing permissions instead of anyone. The Docs API can resolve images the authenticated user has access to.
2. If public access is genuinely needed for the inline image to render, document this clearly in the insertLocalImage tool description so the caller (and ultimately the user) can make an informed choice.

What do you think?

src/index.ts
```diff
@@ -889,2768 +1662,4425 @@ server.setRequestHandler(ListToolsRequestSchema, async () => {
...
-
-      case "getGoogleSheetContent": {
-        const validation = GetGoogleSheetContentSchema.safeParse(request.params.arguments);
-        if (!validation.success) {
-          return errorResponse(validation.error.errors[0].message);
+        const docs = google.docs({ version: 'v1', auth: authClient });
+        const document = await docs.documents.get({ documentId: args.documentId });
+        
+        let content = '';
+        let currentIndex = 1;
```
piotr-agier: The currentIndex variable starts at 1 and is incremented manually as text runs are processed. This will diverge from the actual Google Docs API indices for any document that contains tables, images, inline objects, or other non-paragraph elements because those elements consume index space that this loop doesn't account for.

The actual indices are already available on each element via `textElement.startIndex` / `textElement.endIndex` (which is exactly what `findTextRange` at line 341 uses). Switching to those would make the reported indices reliable for use with the editing tools like `insertText` and `deleteRange`.

src/index.ts
```diff
@@ -889,2768 +1662,4425 @@ server.setRequestHandler(ListToolsRequestSchema, async () => {
...
+            type: 'text',
+            text: `Updated Google Slides presentation with ${args.slides.length} slide(s)\nLink: https://docs.google.com/presentation/d/${args.presentationId}`,
+          }],
+          isError: false,
         };
       }
 
-      case "updateGoogleDoc": {
-        const validation = UpdateGoogleDocSchema.safeParse(request.params.arguments);
+      case "formatGoogleDocText": {
```
piotr-agier: There are now two tools that format text in Google Docs:
- `formatGoogleDocText` (line 4141) - targets by index range only, takes RGB color objects
- `applyTextStyle` (line 5271) - targets by index range or text-find, takes hex color strings, supports more options (`fontFamily`, `backgroundColor`, `linkUrl`)

Same situation with `formatGoogleDocParagraph` (line 4221) vs `applyParagraphStyle` (line 5333).

The apply* versions are strictly more capable. Having both exposed as MCP tools is likely to confuse LLM callers about which to use. Would it make sense to remove the older format* variants?

src/index.ts
```diff
@@ -597,59 +1159,270 @@ const DownloadFileSchema = z.object({
...
+      ensureDriveService();
+      ensureCalendarService();
+    } finally {
+      // Clear the promise after completion (success or failure)
+      authenticationPromise = null;
+    }
+  }
+
+  // If we already have authClient, ensure services are up to date
+  ensureDriveService();
```
piotr-agier, line 1362: `ensureAuthenticated()` unconditionally calls both `ensureDriveService()` and `ensureCalendarService()` on every tool request, even for tools that never touch the calendar (which is most of them).
`ensureDriveService()` also makes a test API call (`drive.about.get`) on every invocation (line 54), so this adds real latency.

A simple fix would be to make these lazy. Call `ensureDriveService()` / `ensureCalendarService()` at the point of use rather than on every request. Or cache them and only recreate when authClient changes.

src/index.ts
```diff
@@ -597,59 +1159,270 @@ const DownloadFileSchema = z.object({
...
+    date: z.string().optional().describe("Date for all-day events (YYYY-MM-DD)"),
+    timeZone: z.string().optional().describe("Time zone (e.g., 'America/Los_Angeles')")
+  }).describe("Start time"),
+  end: z.object({
+    dateTime: z.string().optional().describe("RFC3339 timestamp for timed events"),
+    date: z.string().optional().describe("Date for all-day events (YYYY-MM-DD)"),
+    timeZone: z.string().optional().describe("Time zone (e.g., 'America/Los_Angeles')")
+  }).describe("End time"),
+  attendees: z.array(z.string()).optional().describe("Email addresses of attendees"),
+  sendUpdates: z.enum(["all", "externalOnly", "none"]).optional().default("all").describe("Send notifications to attendees"),
```
piotr-agier: The Zod schema for `createCalendarEvent` declares `.default("all")` for `sendUpdates` (line 1291), meaning after validation `args.sendUpdates` will always be "all" when omitted. But the runtime code at line 5720 applies a `|| 'none'` fallback:
`sendUpdates: args.sendUpdates || 'none'`

Since the Zod default wins, the `|| 'none'` is dead code and the actual behavior is "all", meaning attendees receive email notifications by default. This may surprise users who expect a quiet default. Same issue in `deleteCalendarEvent` (Zod schema line 1320, runtime line 5786) and `updateCalendarEvent` (Zod schema line 1314, runtime line 5765).

I'd suggest picking one consistent default (probably "none" to avoid unexpected emails) and aligning the Zod schema with the runtime code.

src/index.ts
```diff
@@ -889,2768 +1662,4425 @@ server.setRequestHandler(ListToolsRequestSchema, async () => {
...
+    ]
+  };
+});
+
+// -----------------------------------------------------------------------------
+// TOOL CALL REQUEST HANDLER
+// -----------------------------------------------------------------------------
+
+server.setRequestHandler(CallToolRequestSchema, async (request) => {
+  console.error(`[DEBUG] CallTool handler called for tool: ${request.params.name}`);
```
piotr-agier: I'd remove this [DEBUG] message.

src/index.ts
```diff
@@ -889,2768 +1662,4425 @@ server.setRequestHandler(ListToolsRequestSchema, async () => {
...
+  await ensureAuthenticated();
+  console.error(`[DEBUG] After ensureAuthenticated - authClient exists: ${!!authClient}, drive exists: ${!!drive}`);
```
piotr-agier: Same here.

src/index.ts
```diff
@@ -889,2768 +1662,4425 @@ server.setRequestHandler(ListToolsRequestSchema, async () => {
...
+        const sheets = google.sheets({ version: 'v4', auth: authClient });
+        
+        // Parse the range to get sheet ID and grid range
+        const rangeData = await sheets.spreadsheets.get({
+          spreadsheetId: args.spreadsheetId,
+          ranges: [args.range],
+          fields: 'sheets(properties(sheetId,title))'
+        });
+
+        console.error(`[DEBUG] formatGoogleSheetCells - range: ${args.range}`);
```
piotr-agier: I'd remove this [DEBUG] message and the few ones below.

src/auth/server.ts
```diff
@@ -5,14 +5,17 @@ import http from 'http';
...
   'https://www.googleapis.com/auth/drive',
   'https://www.googleapis.com/auth/drive.file',
   'https://www.googleapis.com/auth/drive.readonly',
   'https://www.googleapis.com/auth/documents',
   'https://www.googleapis.com/auth/spreadsheets',
-  'https://www.googleapis.com/auth/presentations'
+  'https://www.googleapis.com/auth/presentations',
+  'https://www.googleapis.com/auth/calendar',
+  'https://www.googleapis.com/auth/calendar.events',
+  'https://www.googleapis.com/auth/script.projects'
```
piotr-agier: Is this to ensure existing Drive tools (search, copy, move, etc.) work correctly with Apps Script project files?

src/index.ts
```diff
@@ -889,2768 +1662,4425 @@ server.setRequestHandler(ListToolsRequestSchema, async () => {
...
-          }],
-          isError: false,
-        };
+        // Build the query string for Google Drive API
+        let queryString = "mimeType='application/vnd.google-apps.document' and trashed=false";
+        if (args.query) {
+          queryString += ` and (name contains '${args.query}' or fullText contains '${args.query}')`;
+        }
+
+        const response = await drive.files.list({
```
piotr-agier, line 5740: This handler doesn't include `supportsAllDrives: true` or `includeItemsFromAllDrives: true` as do `listGoogleSheets` at line 3822. Should we add that here as well?

src/index.ts
ItsiW: Fixed in both listGoogleSheets and listGoogleDocs. Also found the same issue in resolvePath. Will fix in a follow-up.

src/index.ts
ItsiW: The Docs API's `insertInlineImage` fetches the image server-side, not as the authenticated user, so it does need a public URL. Went with option 2 and updated the tool description to make this explicit.

src/index.ts
ItsiW: calendar service is now lazy, and removed the drive.about.get test call

src/index.ts
ItsiW: aligned to "none" everywhere.

src/auth/server.ts
ItsiW: no this was a leftover from some experimentation. I removed it.

src/index.ts
ItsiW: Done. Found a few more calls missing these flags in uploadImageToDriveHelper. I'll fix it in the follow-up

PR #15, opened 2026-01-27
Add uploadFile tool

arahimi-hims:
Adds a new tool `uploadFile` that allows uploading local files to Google Drive.


piotr-agier:
UploadFile already exists in main (tool registration + handler + mime mapping). Unless there’s unique behavior here not present in main, this PR is now redundant. Happy to reopen if there’s a missing capability.

PR #16, opened 2026-01-27
Add addGoogleSheetTab tool

arahimi-hims:
Adds a new tool `addGoogleSheetTab` for adding tabs to Google Sheets.


piotr-agier:
This overlaps with PR #12 (addSpreadsheetSheet). If we merge #12, this becomes redundant; if we keep scope small, this PR is a great focused change. Please confirm any unique behavior so we can decide.

PR #18, opened 2026-01-27
Enhance getGoogleDocContent with formatting info

arahimi-hims:
Updates `getGoogleDocContent` to return detailed formatting information (font, size, style, color) and a font usage summary, using a new `TextSpan` structure.


piotr-agier:
Valuable feature. Two thoughts: (1) output size may get large — consider an `includeFormatting` flag or a separate tool to keep defaults light; (2) this will conflict with PR #20 (multi‑tab). We should coordinate merge order or combine.

PR #30, opened 2026-02-17
feat: Resolve folder paths in search results

zmainen:
## Summary
- Search results now show human-readable folder paths (e.g., `My Drive/Documents/Project`) instead of opaque parent folder IDs
- Recursive path resolution with caching minimizes API calls
- Gracefully falls back to raw folder ID if resolution fails

## Test plan
- [ ] Run a search query and verify paths appear in results
- [ ] Verify caching works (files in the same folder should not duplicate API calls)
- [ ] Verify shared drives resolve correctly

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1535,12 +1535,32 @@ server.setRequestHandler(CallToolRequestSchema, async (request) => {
           q: formattedQuery,
           pageSize: Math.min(pageSize || 50, 100),
           pageToken: pageToken,
-          fields: "nextPageToken, files(id, name, mimeType, modifiedTime, size)",
+          fields: "nextPageToken, files(id, name, mimeType, modifiedTime, size, parents)",
           includeItemsFromAllDrives: true,
           supportsAllDrives: true
         });
 
-        const fileList = res.data.files?.map((f: drive_v3.Schema$File) => `${f.name} (ID: ${f.id}, ${f.mimeType})`).join("\n") || '';
+        // Resolve parent folder IDs to human-readable paths
+        const pathCache: Record<string, string> = {};
+        async function resolveParentPath(folderId: string): Promise<string> {
+          if (pathCache[folderId]) return pathCache[folderId];
+          try {
+            const f = await drive.files.get({ fileId: folderId, fields: "name, parents", supportsAllDrives: true });
+            const parentPath = f.data.parents?.[0] ? await resolveParentPath(f.data.parents[0]) : "";
+            pathCache[folderId] = parentPath ? `${parentPath}/${f.data.name}` : f.data.name!;
+          } catch {
+            pathCache[folderId] = folderId;
+          }
+          return pathCache[folderId];
+        }
+
+        const fileListEntries = await Promise.all(
+          res.data.files?.map(async (f: drive_v3.Schema$File) => {
+            const folderPath = f.parents?.[0] ? await resolveParentPath(f.parents[0]) : "My Drive";
+            return `${f.name} (${f.mimeType}) [id: ${f.id}, path: ${folderPath}]`;
+          }) || []
+        );
+        const fileList = fileListEntries.join("\n");
         log('Search results', { query: userQuery, resultCount: res.data.files?.length });
 
         let response = `Found ${res.data.files?.length ?? 0} files:\n${fileList}`;


src/index.ts
```diff
@@ -1535,12 +1535,32 @@ server.setRequestHandler(CallToolRequestSchema, async (request) => {
...
         });
 
-        const fileList = res.data.files?.map((f: drive_v3.Schema$File) => `${f.name} (ID: ${f.id}, ${f.mimeType})`).join("\n") || '';
+        // Resolve parent folder IDs to human-readable paths
+        const pathCache: Record<string, string> = {};
+        async function resolveParentPath(folderId: string): Promise<string> {
+          if (pathCache[folderId]) return pathCache[folderId];
+          try {
+            const f = await drive.files.get({ fileId: folderId, fields: "name, parents", supportsAllDrives: true });
+            const parentPath = f.data.parents?.[0] ? await resolveParentPath(f.data.parents[0]) : "";
```
piotr-agier, line 1549: This recursion could go deep (or even loop) if the folder hierarchy is unusual: shared drives, shortcuts, or orphaned folders can behave unexpectedly. Would you mind adding a max depth parameter? 

Something like:

```async function resolveParentPath(folderId: string, depth = 0): Promise<string> {
  if (depth > 10 || pathCache[folderId]) return pathCache[folderId] ?? folderId;
  // ...
  const parentPath = f.data.parents?.[0] ? await resolveParentPath(f.data.parents[0], depth + 1) : "";```

File: src/index.ts [hunk 2]
```diff
@@ -1535,12 +1535,32 @@ server.setRequestHandler(CallToolRequestSchema, async (request) => {
...
+            pathCache[folderId] = parentPath ? `${parentPath}/${f.data.name}` : f.data.name!;
+          } catch {
+            pathCache[folderId] = folderId;
+          }
+          return pathCache[folderId];
+        }
+
+        const fileListEntries = await Promise.all(
```
piotr-agier, line 1557: For a search returning 50 results across many folders, this can trigger a lot of extra files.get API calls to walk up each path. The cache helps for siblings, but the first file in each unique folder still needs the full traversal.

Could you make this opt-in (e.g., an includePaths parameter)? That way callers who just need file names and IDs don't pay the overhead, and callers who want paths can request them explicitly.

src/index.ts
```diff
@@ -1535,12 +1535,32 @@ server.setRequestHandler(CallToolRequestSchema, async (request) => {
...
 
-        const fileList = res.data.files?.map((f: drive_v3.Schema$File) => `${f.name} (ID: ${f.id}, ${f.mimeType})`).join("\n") || '';
+        // Resolve parent folder IDs to human-readable paths
+        const pathCache: Record<string, string> = {};
+        async function resolveParentPath(folderId: string): Promise<string> {
+          if (pathCache[folderId]) return pathCache[folderId];
+          try {
+            const f = await drive.files.get({ fileId: folderId, fields: "name, parents", supportsAllDrives: true });
+            const parentPath = f.data.parents?.[0] ? await resolveParentPath(f.data.parents[0]) : "";
+            pathCache[folderId] = parentPath ? `${parentPath}/${f.data.name}` : f.data.name!;
```
piotr-agier, line 1550: f.data.name can technically be undefined in the Drive API types. Since you already have a fallback to `folderId` in the catch block, it'd be consistent to do the same here:


```const name = f.data.name || folderId;
pathCache[folderId] = parentPath ? `${parentPath}/${name}` : name;```

File: src/index.ts [hunk 4]
```diff
@@ -1535,12 +1535,32 @@ server.setRequestHandler(CallToolRequestSchema, async (request) => {
...
+          } catch {
+            pathCache[folderId] = folderId;
+          }
+          return pathCache[folderId];
+        }
+
+        const fileListEntries = await Promise.all(
+          res.data.files?.map(async (f: drive_v3.Schema$File) => {
+            const folderPath = f.parents?.[0] ? await resolveParentPath(f.parents[0]) : "My Drive";
+            return `${f.name} (${f.mimeType}) [id: ${f.id}, path: ${folderPath}]`;
```
piotr-agier, line 1560: Could you use uppercase `ID:` to stay consistent with other tools in the project? For example, listGoogleDocs uses `ID: ${file.id}`.

PR #64, opened 2026-02-28
feat: add paragraph bullets tool, comment resolution, and upload format conversion

nesquena:
This PR adds three enhancements to the Google Drive MCP server:

### 1. New `createParagraphBullets` Tool (Google Docs)

Add or remove bullet points and numbered lists on paragraphs in a Google Doc.

- Supports 13 bullet/numbered style presets (e.g., `BULLET_DISC_CIRCLE_SQUARE`, `NUMBERED_DECIMAL_ALPHA_ROMAN`, `BULLET_CHECKBOX`)
- Use `bulletPreset='NONE'` to remove existing bullets
- Target paragraphs by `startIndex`+`endIndex` or `textToFind`+`matchInstance` (consistent with existing tools like `applyParagraphStyle`)
- Uses the Docs API `createParagraphBullets` and `deleteParagraphBullets` batch requests

### 2. Comment Thread Resolution via `replyToComment`

Added an optional `resolve` boolean parameter to `replyToComment` that resolves the comment thread in a single step — no need for a separate API call after replying.

### 3. Google Workspace Format Conversion on `uploadFile`

Added an optional `convertToGoogleFormat` boolean parameter to `uploadFile` that automatically converts Office files to native Google Workspace formats on upload:

| Source Format | Target Format |
|---|---|
| `.docx` / `.doc` | Google Doc |
| `.xlsx` / `.xls` | Google Sheet |
| `.pptx` / `.ppt` | Google Slides |

File extension is automatically stripped from the name when converting (e.g., `report.docx` → `report`).

---

All changes include corresponding schema validation, tool definitions, and README documentation.

### DIFF
--- a/src/tools/drive.ts
+++ b/src/tools/drive.ts
@@ -10,7 +10,7 @@ import { errorResponse } from '../types.js';
 import { escapeDriveQuery, getMimeTypeFromFilename, TEXT_MIME_TYPES } from '../utils.js';
 import { downloadDriveFile, GOOGLE_WORKSPACE_EXPORT_FORMATS } from '../download-file.js';
 import { getSecureTokenPath } from '../auth/utils.js';
-import { SCOPE_ALIASES, SCOPE_PRESETS, DEFAULT_SCOPES, resolveOAuthScopes } from '../auth/scopes.js';
+import { SCOPE_ALIASES, SCOPE_PRESETS, resolveOAuthScopes } from '../auth/scopes.js';
 
 // ---------------------------------------------------------------------------
 // Constants
@@ -210,6 +210,13 @@ function getGrantedScopesFromAuthClient(ctx: ToolContext): string[] {
   const scopeRaw = ctx.authClient?.credentials?.scope;
   if (!scopeRaw || typeof scopeRaw !== 'string') return [];
   return [...new Set(scopeRaw.split(' ').map((s: string) => s.trim()).filter(Boolean))];
+}
+
+function resolveScopeStatus(ctx: ToolContext): { requestedScopes: string[]; grantedScopes: string[]; missingScopes: string[] } {
+  const requestedScopes = resolveOAuthScopes();
+  const grantedScopes = getGrantedScopesFromAuthClient(ctx);
+  const missingScopes = requestedScopes.filter((s) => !grantedScopes.includes(s));
+  return { requestedScopes, grantedScopes, missingScopes };
 }
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


src/tools/docs.ts
```diff
@@ -1741,6 +1783,77 @@ export async function handleTool(toolName: string, args: Record<string, unknown>
...
+        const range = await findTextRange(
+          ctx,
+          a.documentId,
+          a.textToFind,
+          a.matchInstance || 1
+        );
+        if (!range) {
+          return errorResponse(`Text "${a.textToFind}" not found in document`);
+        }
+        const paraRange = await getParagraphRange(ctx, a.documentId, range.startIndex);
```
piotr-agier: Currently `textToFind` resolves to a single paragraph via `getParagraphRange`, but if the matched text spans multiple paragraphs (or the user selects a broad passage), only the first paragraph gets bullets. This feels like an unnecessary limitation since the `startIndex`+`endIndex` path already supports multi-paragraph ranges. The Docs API handles it natively.

Could we use the matched text range directly instead of collapsing it to one paragraph?

This keeps the code simpler (removes the `getParagraphRange` call entirely) and gives callers richer behavior. Finding a block of text and bulletizing all the paragraphs it touches. If single-paragraph targeting is needed, callers can just match text within one paragraph.

src/tools/drive.ts
```diff
@@ -970,13 +972,31 @@ export async function handleTool(
...
+      const GOOGLE_FORMAT_MAP: Record<string, string> = {
+        'application/vnd.openxmlformats-officedocument.wordprocessingml.document': 'application/vnd.google-apps.document',
+        'application/msword': 'application/vnd.google-apps.document',
+        'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet': 'application/vnd.google-apps.spreadsheet',
+        'application/vnd.ms-excel': 'application/vnd.google-apps.spreadsheet',
+        'application/vnd.openxmlformats-officedocument.presentationml.presentation': 'application/vnd.google-apps.presentation',
+        'application/vnd.ms-powerpoint': 'application/vnd.google-apps.presentation',
+      };
+
+      const targetMimeType = data.convertToGoogleFormat ? GOOGLE_FORMAT_MAP[detectedMime] : undefined;
```
piotr-agier, line 992: When `convertToGoogleFormat: true` is passed but the file's MIME type isn't in `GOOGLE_FORMAT_MAP`, this silently uploads the file without conversion. Since the caller explicitly requested conversion, an early error would be clearer than a silent fallback:

```const targetMimeType = data.convertToGoogleFormat ? GOOGLE_FORMAT_MAP[detectedMime] : undefined;

  if (data.convertToGoogleFormat && !targetMimeType) {
    return errorResponse(
      `Cannot convert MIME type "${detectedMime}" to a Google Workspace format. ` +
      `Supported: .docx, .doc, .xlsx, .xls, .pptx, .ppt`
    );
  }
```

This way the caller finds out immediately rather than discovering later that the file is still in its original format.

src/tools/docs.ts
nesquena: Fixed

src/tools/drive.ts
nesquena: Fixed

nesquena:
@piotr-agier Should be good to go!

PR #67, opened 2026-03-02
fix(docs): Add support for nested tabs to readGoogleDoc and getGoogleDocContent

greyvugrin:
Doc tabs can have an optional `childTabs` array, which can go up to 2 levels (3 levels of tabs deep total).
Example:
Screenshot 2026-03-02 at 1 21 08 PM

`listDocumentTabs` supports the nested tabs.
But when leveraging `readGoogleDoc` with a nested `tabId` supplied, it throws an error (see below).
Screenshot 2026-03-02 at 1 15 52 PM

This PR adds support for this scenario to the `readGoogleDoc` and `getGoogleDocContent` flows.
Additionally, tests were added to cover the various tabbed doc scenarios, and a helper was included to generate docs for each case to reuse them across the tests.

### References:
- https://developers.google.com/workspace/docs/api/reference/rest/v1/documents#Document
- https://developers.google.com/workspace/docs/api/reference/rest/v1/documents#Tab

### DIFF
--- a/src/tools/docs.ts
+++ b/src/tools/docs.ts
@@ -524,6 +524,30 @@ const ApplyParagraphStyleSchema = z.object({
   keepWithNext: z.boolean().optional()
 });
 
+const CreateParagraphBulletsSchema = z.object({
+  documentId: z.string().min(1, "Document ID is required"),
+  startIndex: z.number().int().min(1).optional(),
+  endIndex: z.number().int().min(1).optional(),
+  textToFind: z.string().min(1).optional(),
+  matchInstance: z.number().int().min(1).optional().default(1),
+  bulletPreset: z.enum([
+    'BULLET_DISC_CIRCLE_SQUARE',
+    'BULLET_DIAMONDX_ARROW3D_SQUARE',
+    'BULLET_CHECKBOX',
+    'BULLET_ARROW_DIAMOND_DISC',
+    'BULLET_STAR_CIRCLE_SQUARE',
+    'BULLET_ARROW3D_CIRCLE_SQUARE',
+    'BULLET_LEFTTRIANGLE_DIAMOND_DISC',
+    'NUMBERED_DECIMAL_ALPHA_ROMAN',
+    'NUMBERED_DECIMAL_ALPHA_ROMAN_PARENS',
+    'NUMBERED_DECIMAL_NESTED',
+    'NUMBERED_UPPERALPHA_ALPHA_ROMAN',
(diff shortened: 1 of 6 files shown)


test/integration/docs.test.ts
```diff
@@ -152,6 +260,80 @@ describe('Docs tools', () => {
...
+    });
+    
+    it('reads all tabs including nested when no tabId specified and the document has only one tab with child tabs ', async () => {
+      ctx.mocks.docs.service.documents.get._setImpl(async () => ({
+        data: mockDocs.singleParentNested(),
+      }));
+      const res = await callTool(ctx.client, 'readGoogleDoc', { documentId: 'doc-1' });
+      assert.equal(res.isError, false);
+      // Print out res.content[0].text
+      console.log('Document content:\n', res.content[0].text);
```
piotr-agier: Leftover debug log and comment? Mind removing these two lines?

src/tools/docs.ts
```diff
@@ -205,6 +217,8 @@ async function getParagraphRange(ctx: ToolContext, documentId: string, indexWith
   }
 }
 
+
```
piotr-agier: Nit: extra blank lines introduced here.

src/tools/docs.ts
```diff
@@ -1463,25 +1473,41 @@ export async function handleTool(toolName: string, args: Record<string, unknown>
         return result;
       }
 
+      // Helper to recursively find a tab by ID in the tab tree
```
piotr-agier: Since `collectAllTabsWithLevel` is already at module scope, would it make sense to lift `findTabById` there too?

src/tools/docs.ts
```diff
@@ -1463,25 +1473,41 @@ export async function handleTool(toolName: string, args: Record<string, unknown>
...
+          const tab = findTabById(tabs, a.tabId);
           if (!tab) {
             return errorResponse(`Tab with ID "${a.tabId}" not found. Use listDocumentTabs to see available tabs.`);
           }
           const bodyContent = tab.documentTab?.body?.content;
           if (bodyContent) {
             text = extractText(bodyContent);
           }
-        } else if (tabs.length > 1) {
+        } else if (tabs.length > 1 || tabs[0]?.childTabs) {
```
piotr-agier: In `getGoogleDocContent` you unified the single/multi-tab paths nicely with `collectAllTabsWithLevel + showTabTitles = allTabs.length > 1`. Could the same pattern be used here? That would keep both handlers consistent and avoid the slightly different condition.

greyvugrin: Excellent call.

PR #70, opened 2026-03-03
feat(docs): add comment position context to listComments

ItsiW:
## Summary

- Adds two-tiered comment position context to `listComments` for Google Docs
- **Tier 1**: Search for `quotedFileContent` in doc body via Docs API. If unique, extract surrounding context and exact character offsets. No extra API calls.
- **Tier 2**: For ambiguous matches, export as DOCX and parse `commentRangeStart/End` XML markers. Match DOCX comments to Drive API by `(author, createdTime)`. Extract surrounding context from table rows or paragraphs, and resolve character indices by searching for the context pattern in the flat text.
- Shows full reply chains (author, date, content) instead of just reply count
- Adds `jszip` dependency for DOCX ZIP parsing

### Before
```
1. Alice (1/15/2026) (anchored to: "this")
   Needs more detail here
   └─ 1 reply
   Comment ID: abc123
```

### After
```
1. Alice (1/15/2026)
   Anchored to: "this" [chars 342-346]
   Context before: "We should revisit this"
   Context after: "this section before the deadline"
   Comment: Needs more detail here
   └─ Bob (1/16/2026): Added a paragraph, PTAL
   Comment ID: abc123
```

## How it works

Google Docs returns opaque `kix.xxx` anchor IDs that can't be decoded to positions. This PR works around that:

1. **Tier 1** reads the doc via Docs API, builds a flat text with an offset map, and searches for each comment's `quotedFileContent`. Unique matches get context + exact char indices immediately.

2. **Tier 2** kicks in for ambiguous matches (e.g. a common word appearing many times). It exports the doc as DOCX, which contains explicit `<w:commentRangeStart>` / `<w:commentRangeEnd>` XML markers. These are matched to Drive API comments by `(author, createdTime)`. For table-based docs, row cells are split into before/after context around the commented cell. For non-table docs, surrounding paragraph text is used.

3. **Index resolution** converts DOCX context to Docs API character offsets by searching for the context pattern in the flat text. If the before-context is ambiguous, the after-context (next row or paragraph) is appended to disambiguate.

Resolved comments are excluded from Google's DOCX export, so they only get Tier 1 context (or none if ambiguous).

## Test plan

- [x] Tested on a multi-table doc with 47 comments, including many with identical anchored text
- [x] 100% context accuracy for all active comments
- [x] Character indices resolve correctly when context patterns are unique
- [x] Resolved comments gracefully degrade (no context, no crash)
- [x] Non-table docs fall back to paragraph context extraction
- [x] Doc with no comments returns empty list as before
- [x] Non-Google-Doc file (PDF) returns empty list without crashing

### DIFF
--- a/src/tools/docs.ts
+++ b/src/tools/docs.ts
@@ -424,7 +424,7 @@ async function uploadImageToDriveHelper(
 // ---------------------------------------------------------------------------
 
 /** Context extracted for a single comment (keyed by Drive API comment ID) */
-interface CommentContext {
+export interface CommentContext {
   contextBefore?: string;
   contextAfter?: string;
   startIndex?: number;
@@ -452,7 +452,7 @@ const MAX_PARAGRAPH_CONTEXT_LENGTH = 300;
  * Build flat text from a Google Doc, tracking each character's Docs API startIndex.
  * Handles paragraphs, tables (including nested), and multi-tab docs.
  */
-function buildFlatTextFromDoc(docData: any): FlatTextResult {
+export function buildFlatTextFromDoc(docData: any): FlatTextResult {
   function extractSegments(bodyContent: any[]): TextSegment[] {
     const segs: TextSegment[] = [];
     function fromElements(elements: any[]) {
@@ -506,7 +506,7 @@ function buildFlatTextFromDoc(docData: any): FlatTextResult {
 }
 
 /** Extract cell texts from a DOCX table row XML string */
-function extractRowCells(rowXml: string): string[] {
+export function extractRowCells(rowXml: string): string[] {
   const cells: string[] = [];
   let searchFrom = 0;
   while (true) {
@@ -529,14 +529,14 @@ function extractRowCells(rowXml: string): string[] {
 }
 
 /** DOCX comment info parsed from word/comments.xml */
-interface DocxComment {
+export interface DocxComment {
   author: string;
   date: string;
   content: string;
 }
 
 /** Context extracted from DOCX comment ranges in document.xml */
-interface DocxContextResult {
+export interface DocxContextResult {
   docxComments: Map<number, DocxComment>;
   contextsBefore: Map<number, string>;
   contextsAfter: Map<number, string>;
@@ -547,7 +547,7 @@ interface DocxContextResult {
  * Parse a DOCX export to extract comment positions and surrounding context.
  * Returns DOCX comment metadata and context maps keyed by DOCX comment ID.
  */
-async function resolveContextFromDocx(docxData: ArrayBuffer): Promise<DocxContextResult | null> {
+export async function resolveContextFromDocx(docxData: ArrayBuffer): Promise<DocxContextResult | null> {
   const zip = await JSZip.loadAsync(docxData);
   const commentsXml = await zip.file('word/comments.xml')?.async('string');
   const documentXml = await zip.file('word/document.xml')?.async('string');
@@ -673,7 +673,7 @@ async function resolveContextFromDocx(docxData: ArrayBuffer): Promise<DocxContex
  * Populates the contextMap with matched context. Also resolves Docs API
  * character offsets when flatText/offsetMap are available.
  */
-function matchDocxToDriveComments(
+export function matchDocxToDriveComments(
   driveComments: any[],
   docxResult: DocxContextResult,
   contextMap: Map<string, CommentContext>,
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
         "prepare": "npm run build",
         "test": "npm run test:build && MCP_TESTING=1 node --test .tmp-test/test/**/*.test.js",
         "test:build": "shx rm -rf .tmp-test && tsc -p tsconfig.test.json && shx cp package.json .tmp-test/",
-        "test:unit": "npm run test:build && node --test .tmp-test/test/download-file.test.js .tmp-test/test/utils.test.js",
+        "test:unit": "npm run test:build && node --test .tmp-test/test/download-file.test.js .tmp-test/test/utils.test.js .tmp-test/test/comment-context.test.js",
         "test:integration": "npm run test:build && MCP_TESTING=1 node --test .tmp-test/test/integration/**/*.test.js .tmp-test/test/schema/**/*.test.js",
         "lint": "tsc --noEmit && eslint src/",
         "lint:types": "tsc --noEmit",
--- a/test/comment-context.test.ts
+++ b/test/comment-context.test.ts
@@ -1,0 +1,506 @@
+import assert from 'node:assert/strict';
+import test from 'node:test';
+import JSZip from 'jszip';
+
+import {
+  buildFlatTextFromDoc,
+  extractRowCells,
+  resolveContextFromDocx,
+  matchDocxToDriveComments,
+} from '../src/tools/docs.js';
+
+import type { CommentContext, DocxContextResult } from '../src/tools/docs.js';
+
+// ---------------------------------------------------------------------------
+// buildFlatTextFromDoc
+// ---------------------------------------------------------------------------
+
+test('buildFlatTextFromDoc: extracts paragraphs with correct offsets', () => {
+  const docData = {
+    body: {
+      content: [
+        {
+          paragraph: {
+            elements: [
+              { textRun: { content: 'Hello ' }, startIndex: 1 },
+              { textRun: { content: 'world' }, startIndex: 7 },
+            ],
+          },
+        },
+      ],
+    },
+  };
+  const { flatText, offsetMap } = buildFlatTextFromDoc(docData);
+  assert.equal(flatText, 'Hello world');
+  assert.equal(offsetMap[0], 1);  // 'H' at Docs index 1
+  assert.equal(offsetMap[6], 7);  // 'w' at Docs index 7
+  assert.equal(offsetMap.length, 11);
+});
+
+test('buildFlatTextFromDoc: extracts table cell text', () => {
+  const docData = {
+    body: {
+      content: [
+        {
+          table: {
+            tableRows: [
+              {
+                tableCells: [
+                  {
+                    content: [
+                      {
+                        paragraph: {
+                          elements: [
+                            { textRun: { content: 'Cell A' }, startIndex: 5 },
+                          ],
+                        },
+                      },
+                    ],
+                  },
+                  {
+                    content: [
+                      {
+                        paragraph: {
+                          elements: [
+                            { textRun: { content: 'Cell B' }, startIndex: 20 },
+                          ],
+                        },
+                      },
+                    ],
+                  },
+                ],
+              },
+            ],
+          },
+        },
+      ],
+    },
+  };
+  const { flatText, offsetMap } = buildFlatTextFromDoc(docData);
+  assert.equal(flatText, 'Cell ACell B');
+  assert.equal(offsetMap[0], 5);
+  assert.equal(offsetMap[6], 20);
+});
+
+test('buildFlatTextFromDoc: handles multi-tab docs', () => {
+  const docData = {
+    tabs: [
+      {
+        documentTab: {
+          body: {
+            content: [
+              {
+                paragraph: {
+                  elements: [
+                    { textRun: { content: 'Tab 1' }, startIndex: 1 },
+                  ],
+                },
+              },
+            ],
+          },
+        },
+      },
+      {
+        documentTab: {
+          body: {
+            content: [
+              {
+                paragraph: {
+                  elements: [
+                    { textRun: { content: 'Tab 2' }, startIndex: 10 },
+                  ],
+                },
+              },
+            ],
+          },
+        },
+      },
+    ],
+  };
+  const { flatText } = buildFlatTextFromDoc(docData);
+  assert.equal(flatText, 'Tab 1Tab 2');
+});
+
+test('buildFlatTextFromDoc: returns empty for empty doc', () => {
+  const { flatText, offsetMap } = buildFlatTextFromDoc({});
+  assert.equal(flatText, '');
+  assert.equal(offsetMap.length, 0);
+});
+
+test('buildFlatTextFromDoc: skips elements without startIndex', () => {
+  const docData = {
+    body: {
+      content: [
+        {
+          paragraph: {
+            elements: [
+              { textRun: { content: 'visible' }, startIndex: 0 },
+              { textRun: { content: 'hidden' } }, // no startIndex
+            ],
+          },
+        },
+      ],
+    },
+  };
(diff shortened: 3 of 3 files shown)


src/tools/docs.ts
```diff
@@ -1934,19 +1935,435 @@ export async function handleTool(toolName: string, args: Record<string, unknown>
         };
       }
 
-      // Format comments for display
+      // ── Comment context extraction (two-tiered) ──
```
piotr-agier, line 2272: The main structural concern: this adds ~400 lines of logic inline inside the case "listComments" block. Could you extract the major pieces into standalone functions? Something like:

- `buildFlatTextFromDoc(docData)` - Tier 1 doc parsing + offset map
- `resolveContextFromDocx(docxData, comments)` - Tier 2 DOCX parsing + matching
- `matchDocxToDriveComments(docxComments, driveComments)` - the (author, date) join

This would make each piece independently testable and keep the listComments case readable.

src/tools/docs.ts
```diff
@@ -1934,19 +1935,435 @@ export async function handleTool(toolName: string, args: Record<string, unknown>
...
+                        }
+                      }
+                    }
+                  }
+                  // Remove from map so duplicate timestamps (rare) don't double-match
+                  docxComments.delete(matchedDocxId);
+                }
+              }
+            }
+          } catch {
```
piotr-agier: See comment above.

src/tools/docs.ts
```diff
@@ -1934,19 +1935,435 @@ export async function handleTool(toolName: string, args: Record<string, unknown>
...
+            (comment as any)._endIndex = offsetMap[positions[0] + quoted.length - 1] + 1;
+          } else if (positions.length > 1) {
+            // Ambiguous — need HTML fallback to disambiguate
+            ambiguousComments.push(comment);
+          }
+          // positions.length === 0: quoted text not found (e.g., doc was edited since comment)
+        }
+
+        needsDocxFallback = ambiguousComments.length > 0;
+      } catch {
```
piotr-agier: This bare catch will silently eat auth failures, network timeouts, quota errors, and programming bugs, not just "wrong file type." Two suggestions:

1. For the expected case (non-Doc files), check the MIME type upfront instead of relying on try/catch for control flow.
2. At minimum, catch the error and log it so debugging is possible:

```} catch (err) {
  // Docs API may fail for non-Google-Doc files (Sheets, PDFs, etc.)
  // For genuine errors, log for debugging
  console.error('Tier 1 context extraction failed:', err);
  needsDocxFallback = true;
}
```

Same applies to the Tier 2 `catch {}` at line 2326.

src/tools/docs.ts
```diff
@@ -1934,19 +1935,435 @@ export async function handleTool(toolName: string, args: Record<string, unknown>
...
+            const idx = flatText.indexOf(quoted, searchFrom);
+            if (idx === -1) break;
+            positions.push(idx);
+            searchFrom = idx + 1;
+          }
+
+          if (positions.length === 1) {
+            // Unique match — we know exactly where this comment is
+            const ctx = getContext(positions[0], quoted.length);
+            (comment as any)._contextBefore = ctx.before;
```
piotr-agier: The pattern of mutating the Drive API response objects with `(comment as any)._underscore` properties appears ~20 times and loses all type safety. Consider using a separate `Map<string, CommentContext>` keyed by `comment ID`:

```interface CommentContext {
  contextBefore?: string;
  contextAfter?: string;
  startIndex?: number;
  endIndex?: number;
}
const contextMap = new Map<string, CommentContext>();
```

Then look up `contextMap.get(comment.id)` during formatting. Cleaner, type-safe, and no risk of colliding with future API response fields.

src/tools/docs.ts
```diff
@@ -1934,19 +1935,435 @@ export async function handleTool(toolName: string, args: Record<string, unknown>
...
+            (comment as any)._contextAfter = ctx.after;
+            (comment as any)._contextSource = 'tier1'; // Mark source for index reliability
+            // Store Docs API character offsets (1-based, as used by Docs API)
+            (comment as any)._startIndex = offsetMap[positions[0]];
+            (comment as any)._endIndex = offsetMap[positions[0] + quoted.length - 1] + 1;
```
piotr-agier: If the flat text and quoted text have subtle differences (e.g., Unicode normalization, trailing whitespace), `positions[0] + quoted.length - 1` could exceed `offsetMap.length`, giving `undefined + 1 = NaN`. Worth
adding a bounds check:

```const endIdx = positions[0] + quoted.length - 1;
if (endIdx < offsetMap.length) {
  // ... set _startIndex / _endIndex
}
```

src/tools/docs.ts
```diff
@@ -1934,19 +1935,435 @@ export async function handleTool(toolName: string, args: Record<string, unknown>
...
+
+                // Try table row context first (most comments in table-based docs)
+                const trStart = documentXml.lastIndexOf('<w:tr>', startPos);
+                const trEnd = documentXml.indexOf('</w:tr>', startPos);
+                if (trStart !== -1 && trEnd !== -1 && (startPos - trStart) < 100000) {
+                  const rowXml = documentXml.substring(trStart, trEnd);
+
+                  // Find which cell contains the comment
+                  const cells: { text: string; containsComment: boolean }[] = [];
+                  let cellSearchFrom = 0;
```
piotr-agier: This cell-extraction loop is essentially identical to the `extractRowCells` helper defined at line 2140. Could you reuse the helper here? E.g.:

```
const cellTexts = extractRowCells(rowXml);
```

src/tools/docs.ts
```diff
@@ -1915,10 +1916,10 @@ export async function handleTool(toolName: string, args: Record<string, unknown>
       }
       const a = validation.data;
 
-      // Use Drive API v3 for comments
+      // Use Drive API v3 for comments (include anchor for position data)
       const response = await ctx.getDrive().comments.list({
         fileId: a.documentId,
-        fields: 'comments(id,content,quotedFileContent,author,createdTime,resolved,replies),nextPageToken',
+        fields: 'comments(id,content,quotedFileContent,anchor,author,createdTime,resolved,replies(id,content,author,createdTime)),nextPageToken',
```
piotr-agier: The anchor field was added to the API request, but it's never referenced in the code below. If it's not needed (since you're working around the opaque kix IDs anyway), remove it from the fields list to keep the API response lean.

src/tools/docs.ts
```diff
@@ -1934,19 +1935,435 @@ export async function handleTool(toolName: string, args: Record<string, unknown>
...
+              const rangeStartRegex = /<w:commentRangeStart\s+w:id="(\d+)"\/>/g;
+              let rMatch: RegExpExecArray | null;
+              while ((rMatch = rangeStartRegex.exec(documentXml)) !== null) {
+                const docxId = parseInt(rMatch[1]);
+                const startPos = rMatch.index;
+
+                // Try table row context first (most comments in table-based docs)
+                const trStart = documentXml.lastIndexOf('<w:tr>', startPos);
+                const trEnd = documentXml.indexOf('</w:tr>', startPos);
+                if (trStart !== -1 && trEnd !== -1 && (startPos - trStart) < 100000) {
```
piotr-agier: The 100000 and 50000 thresholds here and at line 2232 aren't explained. What are they guarding against? If they're sanity checks to avoid matching a `<w:tr>` from a completely different table, a short comment would help future readers understand the intent. Named constants would be even better:

```
const MAX_ROW_XML_DISTANCE = 100_000; // Guard against matching rows from distant tables
```

src/tools/docs.ts
```diff
@@ -1934,19 +1935,435 @@ export async function handleTool(toolName: string, args: Record<string, unknown>
...
+                      docxContextsAfter.set(docxId, '');
+                    }
+                    continue;
+                  }
+                }
+
+                // Paragraph fallback for non-table docs
+                const pStart = documentXml.lastIndexOf('<w:p ', startPos);
+                const pEnd = documentXml.indexOf('</w:p>', startPos);
+                if (pStart !== -1 && pEnd !== -1 && (startPos - pStart) < 50000) {
```
piotr-agier: See comment above.

src/tools/docs.ts
```diff
@@ -1934,19 +1935,435 @@ export async function handleTool(toolName: string, args: Record<string, unknown>
...
+            positions.push(idx);
+            searchFrom = idx + 1;
+          }
+
+          if (positions.length === 1) {
+            // Unique match — we know exactly where this comment is
+            const ctx = getContext(positions[0], quoted.length);
+            (comment as any)._contextBefore = ctx.before;
+            (comment as any)._contextAfter = ctx.after;
+            (comment as any)._contextSource = 'tier1'; // Mark source for index reliability
```
piotr-agier: `_contextSource` is written in both Tier 1 and Tier 2 but never appears in the output formatting section. Is this intended for debugging? If so, consider removing it from production code (or exposing it in the output if it's useful for consumers to know the confidence level).

src/tools/docs.ts
```diff
@@ -1934,19 +1935,435 @@ export async function handleTool(toolName: string, args: Record<string, unknown>
...
+          while (true) {
+            const idx = flatText.indexOf(quoted, searchFrom);
+            if (idx === -1) break;
+            positions.push(idx);
+            searchFrom = idx + 1;
+          }
+
+          if (positions.length === 1) {
+            // Unique match — we know exactly where this comment is
+            const ctx = getContext(positions[0], quoted.length);
```
piotr-agier: `ctx` here shadows the outer `ctx: ToolContext` parameter. It compiles fine but is confusing when reading the code, especially since the real `ctx` is used just a few lines above (`ctx.google.docs(...)`). Rename to
something like `matchCtx` or `surrounding`.

src/tools/docs.ts
```diff
@@ -1934,19 +1935,435 @@ export async function handleTool(toolName: string, args: Record<string, unknown>
...
+                const pEnd = documentXml.indexOf('</w:p>', startPos);
+                if (pStart !== -1 && pEnd !== -1 && (startPos - pStart) < 50000) {
+                  const pXml = documentXml.substring(pStart, pEnd);
+                  const pTexts: string[] = [];
+                  const tRegex3 = /<w:t[^>]*>([^<]*)<\/w:t>/g;
+                  let t3: RegExpExecArray | null;
+                  while ((t3 = tRegex3.exec(pXml)) !== null) pTexts.push(t3[1]);
+                  const pText = pTexts.join('').trim();
+                  if (pText) {
+                    docxContextsBefore.set(docxId, pText.length > 300 ? pText.substring(0, 300) + '...' : pText);
```
piotr-agier: Magic number.

src/tools/docs.ts
```diff
@@ -1934,19 +1935,435 @@ export async function handleTool(toolName: string, args: Record<string, unknown>
...
+            // Unique match — we know exactly where this comment is
+            const ctx = getContext(positions[0], quoted.length);
+            (comment as any)._contextBefore = ctx.before;
+            (comment as any)._contextAfter = ctx.after;
+            (comment as any)._contextSource = 'tier1'; // Mark source for index reliability
+            // Store Docs API character offsets (1-based, as used by Docs API)
+            (comment as any)._startIndex = offsetMap[positions[0]];
+            (comment as any)._endIndex = offsetMap[positions[0] + quoted.length - 1] + 1;
+          } else if (positions.length > 1) {
+            // Ambiguous — need HTML fallback to disambiguate
```
piotr-agier: This comment says "HTML fallback" but the actual fallback is DOCX export. Looks like a leftover from an earlier approach.

src/tools/docs.ts
ItsiW: Done — extracted `buildFlatTextFromDoc`, `resolveContextFromDocx`, `matchDocxToDriveComments`, and `extractRowCells` as standalone functions. The listComments case body is now ~130 lines.

src/tools/docs.ts
ItsiW: 👍

src/tools/docs.ts
ItsiW: Done — now checks MIME type upfront via `files.get` and only attempts the Docs API for `application/vnd.google-apps.document`. Both catch blocks now log via `ctx.log()`.

src/tools/docs.ts
ItsiW: Done — using `Map<string, CommentContext>` keyed by comment ID, with a proper interface.

src/tools/docs.ts
ItsiW: 👍 Added bounds check in both Tier 1 and Tier 2 paths.

src/tools/docs.ts
ItsiW: 👍 Now uses the top-level `extractRowCells(rowXml)` for text extraction, with a separate pass to find the comment cell index.

src/tools/docs.ts
ItsiW: 👍 Removed.

src/tools/docs.ts
ItsiW: 👍 Added `MAX_ROW_XML_DISTANCE`, `MAX_PARAGRAPH_XML_DISTANCE`, and `MAX_PARAGRAPH_CONTEXT_LENGTH` as named constants with comments.

src/tools/docs.ts
ItsiW: 👍

src/tools/docs.ts
ItsiW: 👍 Removed entirely — was only useful for debugging during development.

src/tools/docs.ts
ItsiW: 👍 Renamed to `surrounding`.

src/tools/docs.ts
ItsiW: 👍

src/tools/docs.ts
ItsiW: 👍 Fixed.

src/tools/docs.ts
```diff
@@ -1974,7 +2439,7 @@ export async function handleTool(toolName: string, args: Record<string, unknown>
       const response = await ctx.getDrive().comments.get({
         fileId: a.documentId,
         commentId: a.commentId,
-        fields: 'id,content,quotedFileContent,author,createdTime,resolved,replies(id,content,author,createdTime)'
+        fields: 'id,content,quotedFileContent,anchor,author,createdTime,resolved,replies(id,content,author,createdTime)'
```
piotr-agier: Leftover: anchor was removed from `listComments` but is still in the `getComment` fields here. Since it's not used in either path, can you remove it for consistency?

ItsiW:
@piotr-agier just a heads up that google just released https://github.com/googleworkspace/cli so it might make sense to switch over to that after it stabilizes. Appreciate all the work!

PR #95, opened 2026-04-08
Add Slides element management tools

timlantin:
## Summary

Adds 5 new Google Slides tools for programmatic slide layout:

**Image insertion:**
- `insertSlidesImageFromUrl` — insert an image from a public URL into a slide
- `insertSlidesLocalImage` — upload a local image to Drive and insert into a slide

**Element management:**
- `moveSlideElement` — reposition and/or resize any page element (images, text boxes, shapes) using transform updates with scale-based resizing
- `deleteSlideElement` — remove elements from slides
- `getSlideElementInfo` — diagnostic tool that returns actual rendered bounds (intrinsic size × scale + translate) for all elements on a slide, with off-page detection

These tools enable full programmatic control over slide layout — useful for building presentations with precise figure placement while preserving aspect ratios.

## Test plan
- [x] TypeScript typecheck passes
- [x] Bundle builds successfully
- [x] Tested against live Google Slides API: insert, move, resize, delete, and inspect operations verified on a real presentation
- [x] Off-page detection correctly flags elements extending past slide boundaries

### DIFF
--- a/src/tools/slides.ts
+++ b/src/tools/slides.ts
@@ -1,5 +1,8 @@
 import { z } from 'zod';
 import { v4 as uuidv4 } from 'uuid';
+import { existsSync } from 'fs';
+import { createReadStream } from 'fs';
+import { basename, extname } from 'path';
 import type { ToolDefinition, ToolResult, ToolContext } from '../types.js';
 import { errorResponse } from '../types.js';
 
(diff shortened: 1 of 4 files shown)


piotr-agier:
Could you update the tool registry test that's failing @timlantin

src/tools/slides.ts
```diff
@@ -154,6 +157,155 @@ const ExportSlideThumbnailSchema = z.object({
...
+  width: z.number().optional(),
+  height: z.number().optional(),
+  makePublic: z.boolean().optional().default(true),
+});
+
+// ---------------------------------------------------------------------------
+// Helpers
+// ---------------------------------------------------------------------------
+
+async function uploadImageToDriveForSlides(
```
piotr-agier: This function is a near-duplicate of uploadImageToDriveHelper in src/tools/docs.ts:367-446. Same MIME map, same upload/permission/webContentLink flow. The only meaningful differences are that the docs version supports a parentFolderId and defaults makePublic to false.

Could we extract one shared helper (e.g. src/utils/driveImageUpload.ts) that accepts both params, and have both docs and slides call into it? Future fixes (new MIME types, error handling improvements) should only need to happen once.

src/tools/slides.ts
```diff
@@ -154,6 +157,155 @@ const ExportSlideThumbnailSchema = z.object({
...
+
+const InsertSlidesLocalImageSchema = z.object({
+  presentationId: z.string().min(1, "Presentation ID is required"),
+  pageObjectId: z.string().min(1, "Slide/page object ID is required"),
+  localImagePath: z.string().min(1, "Local image path is required"),
+  x: z.number().optional().default(0),
+  y: z.number().optional().default(0),
+  width: z.number().optional(),
+  height: z.number().optional(),
+  makePublic: z.boolean().optional().default(true),
```
piotr-agier: Defaulting makePublic to true leaves every locally-uploaded image permanently world-readable, which is a sharper default than I'd expect from a tool that looks like it just inserts an image.

The Slides API only needs the URL at the moment of batchUpdate. Once the image is embedded, the original Drive file's public link isn't needed anymore. Could we revoke the anyone permission after insertImageIntoSlide returns successfully?

src/tools/slides.ts
```diff
@@ -1482,6 +1709,150 @@ export async function handleTool(
...
+    }
+
+    case "insertSlidesImageFromUrl": {
+      const validation = InsertSlidesImageFromUrlSchema.safeParse(args);
+      if (!validation.success) return errorResponse(validation.error.errors[0].message);
+      const a = validation.data;
+      return insertImageIntoSlide(ctx, a.presentationId, a.pageObjectId, a.imageUrl, a.x, a.y, a.width, a.height);
+    }
+
+    case "insertSlidesLocalImage": {
```
piotr-agier, line 1808: After the image is embedded in the slide, the uploaded Drive file is no longer referenced by anything. Slides stores its own copy. Right now every call to this tool leaves an orphaned image in the user's Drive root, which will accumulate silently over time.

Can we simply delete the intermediary file after a successful insertImageIntoSlide?

src/tools/slides.ts
```diff
@@ -1482,6 +1709,150 @@ export async function handleTool(
...
+    case "moveSlideElement": {
+      const validation = MoveSlideElementSchema.safeParse(args);
+      if (!validation.success) return errorResponse(validation.error.errors[0].message);
+      const a = validation.data;
+
+      const slidesService = ctx.google.slides({ version: 'v1', auth: ctx.authClient });
+
+      // First get the current element to read its existing transform and size
+      const pres = await slidesService.presentations.get({ presentationId: a.presentationId });
+      let currentTransform: any = null;
```
piotr-agier: The googleapis SDK exposes exact types for these: slides_v1.Schema$AffineTransform, slides_v1.Schema$Size, and slides_v1.Schema$Request[]. Using them here would give us compile-time protection against typos in field names (shearX, translateY, etc.) which matters for this handler specifically since it's constructing a transform by hand from previously-read values.

I know `any` shows up elsewhere in this file, but I'd rather not grow the footprint in new code when the typed alternatives are a one-line change.

src/tools/slides.ts
```diff
@@ -1482,6 +1709,150 @@ export async function handleTool(
...
+        newScaleX = a.width / origWidth;
+      }
+      if (a.height !== undefined) {
+        newScaleY = a.height / origHeight;
+      }
+
+      const newX = a.x ?? (currentTransform.translateX || 0);
+      const newY = a.y ?? (currentTransform.translateY || 0);
+
+      const requests: any[] = [{
```
piotr-agier: See comment on line 1760.

src/tools/slides.ts
```diff
@@ -154,6 +157,155 @@ const ExportSlideThumbnailSchema = z.object({
...
+  imageUrl: string,
+  x: number,
+  y: number,
+  width?: number,
+  height?: number,
+): Promise<ToolResult> {
+  const slidesService = ctx.google.slides({ version: 'v1', auth: ctx.authClient });
+  const objectId = `img_${uuidv4().substring(0, 8)}`;
+
+  const elementProperties: any = {
```
piotr-agier: slides_v1.Schema$PageElementProperties is the right type here and would catch mistakes in the size / transform shapes at compile time.

src/tools/slides.ts
```diff
@@ -1482,6 +1709,150 @@ export async function handleTool(
       };
     }
 
+    case "getSlideElementInfo": {
+      const validation = GetSlideElementInfoSchema.safeParse(args);
+      if (!validation.success) return errorResponse(validation.error.errors[0].message);
+      const a = validation.data;
+
+      const slidesService = ctx.google.slides({ version: 'v1', auth: ctx.authClient });
+      const pres = await slidesService.presentations.get({ presentationId: a.presentationId });
```
piotr-agier: We're calling `presentations.get({ presentationId })` with no field mask, which pulls the full presentation, including every slide, every element, every text run, just to read one element's transform and size. For
large decks that's a lot of wasted payload.

In `getSlideElementInfo`, when `slideObjectId` is provided, use `presentations.pages.get({ presentationId, pageObjectId })`. The SDK exposes this method.

src/tools/slides.ts
```diff
@@ -1482,6 +1709,150 @@ export async function handleTool(
...
+
+    case "moveSlideElement": {
+      const validation = MoveSlideElementSchema.safeParse(args);
+      if (!validation.success) return errorResponse(validation.error.errors[0].message);
+      const a = validation.data;
+
+      const slidesService = ctx.google.slides({ version: 'v1', auth: ctx.authClient });
+
+      // First get the current element to read its existing transform and size
+      const pres = await slidesService.presentations.get({ presentationId: a.presentationId });
```
piotr-agier: Similar as in line 1718. Add a field mask: `fields: 'slides(pageElements(objectId,transform,size))'`.

src/tools/slides.ts
```diff
@@ -1482,6 +1709,150 @@ export async function handleTool(
...
+          const intrH = s.height?.magnitude || 0;
+          const scX = t.scaleX || 1;
+          const scY = t.scaleY || 1;
+          const tx = t.translateX || 0;
+          const ty = t.translateY || 0;
+          const renderedW = intrW * scX;
+          const renderedH = intrH * scY;
+          const right = tx + renderedW;
+          const bottom = ty + renderedH;
+          const offPage = (right > slideWidth || bottom > slideHeight) ? ' *** OFF PAGE ***' : '';
```
piotr-agier: This catches elements that overflow to the right or bottom, but misses the symmetric case: an element with translateX: -500000 sits partially off the left edge but won't be flagged. Same for negative Y.

```
const offPage = (tx < 0 || ty < 0 || right > slideWidth || bottom > slideHeight) ? ' *** OFF PAGE ***' : '';
```

src/tools/slides.ts
```diff
@@ -1482,6 +1709,150 @@ export async function handleTool(
...
+        },
+      }];
+
+      await slidesService.presentations.batchUpdate({
+        presentationId: a.presentationId,
+        requestBody: { requests },
+      });
+
+      return {
+        content: [{ type: 'text', text: `Moved element ${a.objectId} to (${newX}, ${newY})` }],
```
piotr-agier: When the caller passes width and/or height, the element is also resized, but the response still only says "Moved". Something like:

```
const didResize = a.width !== undefined || a.height !== undefined;
const action = didResize ? 'Moved/resized' : 'Moved';
```

piotr-agier (changes requested):

piotr-agier (approved):

PR #98, opened 2026-04-15
feat: configurable OAuth auth server port via GOOGLE_DRIVE_MCP_AUTH_PORT

thatjpcsguy:
## Summary

- Adds `GOOGLE_DRIVE_MCP_AUTH_PORT` environment variable to configure the starting port for the OAuth callback server
- Default behavior unchanged (ports 3000–3004)
- When set, the server tries 5 consecutive ports starting from the configured value (e.g., `GOOGLE_DRIVE_MCP_AUTH_PORT=3100` → tries 3100–3104)

## Motivation

Port 3000 is commonly used by dev servers (React, Next.js, Rails, etc.), causing frequent conflicts when authenticating. This is a one-line change to the `AuthServer` constructor that reads from an env var with a fallback to the existing default.

## Changes

- `src/auth/server.ts`: Read `GOOGLE_DRIVE_MCP_AUTH_PORT` env var in constructor
- `README.md`: Document the new env var in configuration and troubleshooting sections

## Usage

```bash
export GOOGLE_DRIVE_MCP_AUTH_PORT=3100
npx @piotr-agier/google-drive-mcp auth
```

### DIFF
--- a/src/auth/server.ts
+++ b/src/auth/server.ts
@@ -14,7 +14,7 @@ export class AuthServer {
   private app: express.Express;
   private server: http.Server | null = null;
   private tokenManager: TokenManager;
-  private portRange: { start: number; end: number };
+  public readonly portRange: { start: number; end: number };
   public authCompletedSuccessfully = false; // Flag for standalone script
 
   constructor(oauth2Client: OAuth2Client) {
--- a/src/auth/server.ts
+++ b/src/auth/server.ts
@@ -21,7 +21,13 @@ export class AuthServer {
     this.baseOAuth2Client = oauth2Client;
     this.tokenManager = new TokenManager(oauth2Client);
     this.app = express();
-    const portStart = parseInt(process.env.GOOGLE_DRIVE_MCP_AUTH_PORT || '3000', 10);
+    const raw = process.env.GOOGLE_DRIVE_MCP_AUTH_PORT;
+    const portStart = raw ? Number(raw) : 3000;
+    if (!Number.isInteger(portStart) || portStart < 1 || portStart > 65531) {
+      throw new Error(
+        `Invalid GOOGLE_DRIVE_MCP_AUTH_PORT: "${raw}". Must be an integer between 1 and 65531.`
+      );
+    }
     this.portRange = { start: portStart, end: portStart + 4 };
     this.setupRoutes();
   }
--- a/src/auth/server.ts
+++ b/src/auth/server.ts
@@ -21,7 +21,8 @@ export class AuthServer {
     this.baseOAuth2Client = oauth2Client;
     this.tokenManager = new TokenManager(oauth2Client);
     this.app = express();
-    this.portRange = { start: 3000, end: 3004 };
+    const portStart = parseInt(process.env.GOOGLE_DRIVE_MCP_AUTH_PORT || '3000', 10);
+    this.portRange = { start: portStart, end: portStart + 4 };
     this.setupRoutes();
   }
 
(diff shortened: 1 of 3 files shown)


src/auth/server.ts
```diff
@@ -21,7 +21,8 @@ export class AuthServer {
     this.baseOAuth2Client = oauth2Client;
     this.tokenManager = new TokenManager(oauth2Client);
     this.app = express();
-    this.portRange = { start: 3000, end: 3004 };
+    const portStart = parseInt(process.env.GOOGLE_DRIVE_MCP_AUTH_PORT || '3000', 10);
```
piotr-agier: `parseInt('abc', 10)` returns `NaN`, which silently propagates into `portRange` and makes the bind loop in `startServerOnAvailablePort` (line 167) never iterate. The user then sees:

> No available ports for authentication server (tried ports NaN - NaN)

The same class of problem hits negative numbers and values above the valid port range. I'd rather we fail fast with a message that points at the actual problem:

```
const raw = process.env.GOOGLE_DRIVE_MCP_AUTH_PORT;
const portStart = raw ? Number(raw) : 3000;
if (!Number.isInteger(portStart) || portStart < 1 || portStart > 65531) {
  throw new Error(
    `Invalid GOOGLE_DRIVE_MCP_AUTH_PORT: "${raw}". Must be an integer between 1 and 65531.`
  );
}
this.portRange = { start: portStart, end: portStart + 4 };
```
(Upper bound leaves room for the +4. Using Number() over parseInt also rejects "3000abc" rather than silently accepting it as 3000.)

piotr-agier:
In `src/index.ts:424`

This hardcoded range is now stale when the env var is set "Authentication failed. Could not start server or validate existing tokens. Check port availability (3000-3004) and try again."

If someone set `GOOGLE_DRIVE_MCP_AUTH_PORT=3100`, this message will send them looking at the wrong ports. Two easy options:

1. Expose the range from AuthServer and interpolate it here, e.g. `Check port availability (${authServerInstance.portRange.start}-${authServerInstance.portRange.end}) and try again.`

2. Or make it generic: "…Check port availability and try again."

The first is more helpful to the user.

piotr-agier:
Please add `GOOGLE_DRIVE_MCP_AUTH_PORT` to the --help output

The new variable is documented in the README but not listed here alongside GOOGLE_DRIVE_MCP_TOKEN_PATH. Users discovering options via --help won't see it.

Suggest adding a line right below the existing `GOOGLE_DRIVE_MCP_TOKEN_PATH` entry:

GOOGLE_DRIVE_MCP_AUTH_PORT            Starting port for OAuth callback server (default: 3000, uses 5 consecutive ports)

piotr-agier:
Missed one doc location: the reference table.

The structured Optional (for customization) reference table a few hundred lines down is the one most readers scan, and it's still missing the new variable.

Suggest adding a row after `GOOGLE_DRIVE_MCP_TOKEN_PATH`:

| `GOOGLE_DRIVE_MCP_AUTH_PORT` | Starting port for OAuth callback server (uses 5 consecutive ports) | `3000` | `3100` |

thatjpcsguy:
Thanks for the review! Done in 91c7732 — `portRange` is now `public readonly` on `AuthServer` and the failure message in `src/index.ts` interpolates `${start}-${end}`, so the hint tracks the configured range.

> Please add GOOGLE_DRIVE_MCP_AUTH_PORT to the --help output

Thanks — added to the --help output in https://github.com/piotr-agier/google-drive-mcp/commit/91c773234f71897bed8171327d519973dc8df28d, right below GOOGLE_DRIVE_MCP_TOKEN_PATH using the suggested wording.

> Missed one doc location: the reference table.

Good catch — added the row to the Optional (for customization) reference table in 91c7732. Thanks for the thorough review!

src/auth/server.ts
thatjpcsguy: Thanks, great catch. Applied your snippet almost verbatim in 430cf2f — `Number()` + `Number.isInteger()` with the 1–65531 bound, throwing immediately with the variable name and the invalid value so the user sees the real problem instead of the `NaN - NaN` port-loop message.

PR #100, opened 2026-04-18
feat: add emailMessage support to addPermission and shareFile

### LINKED ISSUE #99: Add emailMessage support to addPermission and shareFile
## Summary

The Google Drive API supports an `emailMessage` parameter on `permissions.create` that lets you include a custom message in the sharing notification email. This is the same "Add a message" field users see in the Google Drive sharing UI.

Currently, `addPermission` and `shareFile` both call `permissions.create` but don't expose the `emailMessage` parameter. This means there's no way to include a custom note when sharing — the recipient just gets a generic "X shared a file with you" email.

## What needs to change

Three layers per tool (`addPermission` + `shareFile`):

1. **Zod schema** — add `emailMessage: z.string().optional()`
2. **Tool input schema** (the JSON definition) — add `emailMessage: { type: "string", description: "Custom message to include in the notification email" }`
3. **API call** — pass `emailMessage` through to `permissions.create` when provided

## Google Drive API reference

[`permissions.create` — `emailMessage` parameter](https://developers.google.com/drive/api/reference/rest/v3/permissions/create)

> A plain text custom message to include in the notification email.

## Context

We patched this locally for our use case and it works. Happy to open a PR if that's preferred.

hopp2it:
## Summary

- Adds `emailMessage` optional parameter to `addPermission` and `shareFile` tools
- Wires it through at all three layers: Zod validation schema, MCP tool input schema, and the `permissions.create` API call
- Mirrors the "Add a message" field in the Google Drive sharing UI

The Google Drive API has always supported [`emailMessage` on `permissions.create`](https://developers.google.com/drive/api/reference/rest/v3/permissions/create) — this just surfaces it through the MCP server.

## Changes

One file, 8 insertions, 2 deletions in `src/tools/drive.ts`:

1. `AddPermissionSchema` — added `emailMessage: z.string().optional()`
2. `ShareFileSchema` — added `emailMessage: z.string().optional()`
3. `addPermission` tool input schema — added `emailMessage` property
4. `shareFile` tool input schema — added `emailMessage` property
5. `addPermission` handler — passes `emailMessage` to `permissions.create`
6. `shareFile` handler — passes `emailMessage` to `permissions.create`

## Test

- TypeScript compiles clean (`npm run typecheck`)
- Build passes (`npm run build`)
- Tested locally: sharing a file with `emailMessage` sends the custom message in the notification email

Closes #99

### DIFF
--- a/src/tools/drive.ts
+++ b/src/tools/drive.ts
@@ -423,7 +423,7 @@ export const toolDefinitions: ToolDefinition[] = [
         role: { type: "string", enum: ["owner", "organizer", "fileOrganizer", "writer", "commenter", "reader"], description: "Permission role" },
         type: { type: "string", enum: ["user", "group", "domain", "anyone"], description: "Principal type" },
         sendNotificationEmail: { type: "boolean", description: "Send notification email" },
-        emailMessage: { type: "string", description: "Custom message to include in the notification email" }
+        emailMessage: { type: "string", description: "Custom message to include in the notification email. Ignored unless sendNotificationEmail is true." }
       },
       required: ["fileId", "emailAddress"]
     }
@@ -464,7 +464,7 @@ export const toolDefinitions: ToolDefinition[] = [
         emailAddress: { type: "string", description: "User email" },
         role: { type: "string", enum: ["writer", "commenter", "reader"], description: "Access role" },
         sendNotificationEmail: { type: "boolean", description: "Send notification email" },
-        emailMessage: { type: "string", description: "Custom message to include in the notification email" }
+        emailMessage: { type: "string", description: "Custom message to include in the notification email. Ignored unless sendNotificationEmail is true." }
       },
       required: ["fileId", "emailAddress"]
     }
--- a/src/tools/drive.ts
+++ b/src/tools/drive.ts
@@ -135,6 +135,7 @@ const AddPermissionSchema = z.object({
   role: z.enum(["owner", "organizer", "fileOrganizer", "writer", "commenter", "reader"]).default("reader"),
   type: z.enum(["user", "group", "domain", "anyone"]).default("user"),
   sendNotificationEmail: z.boolean().optional().default(false),
+  emailMessage: z.string().optional(),
 });
 
 const UpdatePermissionSchema = z.object({
@@ -158,6 +159,7 @@ const ShareFileSchema = z.object({
   emailAddress: z.string().email("Valid email is required"),
   role: z.enum(["writer", "commenter", "reader"]).default("reader"),
   sendNotificationEmail: z.boolean().optional().default(true),
+  emailMessage: z.string().optional(),
 });
 
 const ConvertPdfToGoogleDocSchema = z.object({
@@ -420,7 +422,8 @@ export const toolDefinitions: ToolDefinition[] = [
         emailAddress: { type: "string", description: "Target user/group email" },
         role: { type: "string", enum: ["owner", "organizer", "fileOrganizer", "writer", "commenter", "reader"], description: "Permission role" },
         type: { type: "string", enum: ["user", "group", "domain", "anyone"], description: "Principal type" },
-        sendNotificationEmail: { type: "boolean", description: "Send notification email" }
+        sendNotificationEmail: { type: "boolean", description: "Send notification email" },
+        emailMessage: { type: "string", description: "Custom message to include in the notification email" }
       },
       required: ["fileId", "emailAddress"]
     }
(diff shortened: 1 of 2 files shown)


src/tools/drive.ts
```diff
@@ -420,7 +422,8 @@ export const toolDefinitions: ToolDefinition[] = [
         emailAddress: { type: "string", description: "Target user/group email" },
         role: { type: "string", enum: ["owner", "organizer", "fileOrganizer", "writer", "commenter", "reader"], description: "Permission role" },
         type: { type: "string", enum: ["user", "group", "domain", "anyone"], description: "Principal type" },
-        sendNotificationEmail: { type: "boolean", description: "Send notification email" }
+        sendNotificationEmail: { type: "boolean", description: "Send notification email" },
+        emailMessage: { type: "string", description: "Custom message to include in the notification email" }
```
piotr-agier: The Drive API silently ignores `emailMessage` unless `sendNotificationEmail` is `true` — and for `addPermission`, `sendNotificationEmail` defaults to `false` (line 137). A caller that sets only `emailMessage` will see no error and no email, which is a confusing failure mode given the description implies the message will be delivered.

Since MCP clients surface this description verbatim, worth tightening it here:

```diff
-        emailMessage: { type: "string", description: "Custom message to include in the notification email" }
+        emailMessage: { type: "string", description: "Custom message to include in the notification email. Ignored unless sendNotificationEmail is true." }
```

Same wording would apply to `shareFile` at line 467. `shareFile` defaults `sendNotificationEmail` to `true` so the footgun is smaller there, but the constraint is identical and worth stating consistently.

piotr-agier (commented):
Two non-blocking suggestions — overall this looks good, just a clarity nit and a test gap.

**Add a mock-level assertion that `emailMessage` reaches `permissions.create`.** Since the field is only observable via a real email send, nothing currently prevents a future refactor from silently dropping it. The existing `tracker.getCalls('permissions.create')` pattern at `test/integration/drive.test.ts:371` is a clean fit — one test per tool, plus a negative case confirming the key is absent when not provided, would lock in the wiring cheaply.

```ts
it('addPermission forwards emailMessage', async () => {
  const res = await callTool(ctx.client, 'addPermission', {
    fileId: 'file-1', emailAddress: 'user@example.com', role: 'reader', type: 'user',
    sendNotificationEmail: true, emailMessage: 'Welcome!',
  });
  assert.equal(res.isError, false);
  const [call] = ctx.mocks.drive.tracker.getCalls('permissions.create');
  assert.equal(call.args[0].emailMessage, 'Welcome!');
});

it('shareFile forwards emailMessage', async () => {
  const res = await callTool(ctx.client, 'shareFile', {
    fileId: 'file-1', emailAddress: 'user@example.com', role: 'writer',
    emailMessage: 'Sharing this with you',
  });
  assert.equal(res.isError, false);
  const [call] = ctx.mocks.drive.tracker.getCalls('permissions.create');
  assert.equal(call.args[0].emailMessage, 'Sharing this with you');
});

it('shareFile omits emailMessage when not provided', async () => {
  const res = await callTool(ctx.client, 'shareFile', {
    fileId: 'file-1', emailAddress: 'user@example.com', role: 'writer',
  });
  assert.equal(res.isError, false);
  const [call] = ctx.mocks.drive.tracker.getCalls('permissions.create');
  assert.equal('emailMessage' in call.args[0], false);
});
```

src/tools/drive.ts
hopp2it: Fixed in e3d4612. Added "Ignored unless sendNotificationEmail is true." to both `addPermission` (line 426) and `shareFile` (line 467).

hopp2it:
Both suggestions addressed in e3d4612.

**Test additions** (test/integration/drive.test.ts:377-407): Three tests added following the existing tracker pattern.

- `addPermission forwards emailMessage to permissions.create` — asserts the field reaches the API call
- `shareFile forwards emailMessage to permissions.create` — same, via shareFile's create path
- `shareFile omits emailMessage when not provided` — negative case, confirms the key is absent

One implementation note: `shareFile` calls `permissions.list` first for idempotent update behavior, and the default mock returns an existing `user@example.com` permission — which routes through the update path (no create call). The two shareFile tests stub `permissions.list` to return empty so the create branch runs. The addPermission test doesn't need this since addPermission goes straight to create.

Full suite: 60/60 drive tests pass (270/271 overall; the one unrelated failure is a pre-existing Windows-specific `--port 0` exit-code flake in `cli-args.test.js`).

PR #107, opened 2026-04-21
feat(auth): support domain-wide delegation via GOOGLE_DRIVE_MCP_SUBJECT

arahimi-hims:
## Summary

Add Workspace domain-wide delegation (DWD) to service-account mode so an admin-authorized SA can act *as a specific user* — not just as itself.

Concretely:

- Read `GOOGLE_DRIVE_MCP_SUBJECT` and pass it as `clientOptions.subject` to `GoogleAuth`. This is the documented DWD impersonation path for JWT-based auth: https://developers.google.com/workspace/guides/create-credentials#service-account
- Honor `GOOGLE_DRIVE_MCP_SCOPES` in SA mode too (previously only the interactive OAuth path used it), so admins can narrow the JWT to a subset of the authorized scopes.
- Extract `buildServiceAccountAuthOptions()` so the subject + scopes resolution can be unit-tested without touching the filesystem.

## Why

Some Google APIs refuse requests from a pure service-account identity — Drive reads scoped to a user's "My Drive", Calendar ACL writes against a personal calendar, Docs created in a specific user's namespace, etc. DWD is the supported way around this: the Workspace admin grants the SA's client ID permission to impersonate users under specific scopes, and the SA mints JWTs with `sub=<user>` to act as them.

Before this change the MCP required OAuth refresh tokens for user-identity workflows, which forces operators to attach a personal Google account to every deployment. With `GOOGLE_DRIVE_MCP_SUBJECT` they can instead stand up a dedicated bot seat, delegate the SA to impersonate it, and drop in a single SA key.

## Behavior

- `GOOGLE_APPLICATION_CREDENTIALS` alone → unchanged (SA acts as itself).
- `GOOGLE_APPLICATION_CREDENTIALS` + `GOOGLE_DRIVE_MCP_SUBJECT=user@example.com` → SA impersonates that user. Workspace admin must authorize the SA's client ID for the requested scopes under Security → API controls → Manage Domain-wide Delegation. A startup log line now notes the impersonation target so operators can confirm which identity is being used.
- `GOOGLE_DRIVE_MCP_SCOPES` → resolved the same way as in OAuth mode (aliases + full URLs).

## Test plan

- [x] `npm run build` passes (tsc + bundle)
- [x] `npm test` passes — 279/279 tests, 18/18 in `external-auth.test.ts`
- [x] Five new unit tests for `buildServiceAccountAuthOptions`: default scopes, subject present, whitespace-trimmed subject, blank subject omitted, `GOOGLE_DRIVE_MCP_SCOPES` override
- [ ] End-to-end: verified locally against a Workspace tenant (tpm-agent@ bot seat, drive.file + documents + spreadsheets + presentations scopes) — can attach a session transcript if useful

## Notes

- Fully backwards compatible. Existing SA-mode deployments with no `GOOGLE_DRIVE_MCP_SUBJECT` set behave identically.
- Help text in `src/index.ts` updated to document both new env vars.
- Happy to adjust the env-var name if you'd prefer a different prefix (e.g. `GOOGLE_DRIVE_MCP_IMPERSONATE`).

### DIFF
--- a/src/auth/externalAuth.ts
+++ b/src/auth/externalAuth.ts
@@ -3,8 +3,8 @@
 // ---------------------------------------------------------------------------
 
 import { OAuth2Client } from 'google-auth-library';
-import { GoogleAuth } from 'google-auth-library';
-import { DEFAULT_SCOPES } from './scopes.js';
+import { GoogleAuth, GoogleAuthOptions } from 'google-auth-library';
+import { resolveOAuthScopes } from './scopes.js';
 
 // ---------------------------------------------------------------------------
 // Service Account mode
@@ -16,18 +16,48 @@ export function isServiceAccountMode(): boolean {
 }
 
 /**
+ * Build `GoogleAuth` options from the current environment.
+ *
+ * When `GOOGLE_DRIVE_MCP_SUBJECT` is set, the returned options include
+ * `clientOptions.subject`, which instructs `GoogleAuth` to mint a JWT that
+ * impersonates the given user via domain-wide delegation. The Workspace admin
+ * must have authorized the service account's client ID for the requested
+ * scopes under Security → API controls → Manage Domain-wide Delegation.
+ *
+ * Scopes are resolved via {@link resolveOAuthScopes}, so `GOOGLE_DRIVE_MCP_SCOPES`
+ * narrows the SA's authority the same way it does for interactive OAuth.
+ *
+ * Exported so tests can assert the shape without hitting the filesystem.
+ */
+export function buildServiceAccountAuthOptions(): GoogleAuthOptions {
+  const keyFile = process.env.GOOGLE_APPLICATION_CREDENTIALS!;
+  const subject = process.env.GOOGLE_DRIVE_MCP_SUBJECT?.trim();
+
+  const options: GoogleAuthOptions = {
+    keyFile,
+    scopes: resolveOAuthScopes(),
+  };
+
+  if (subject) {
+    options.clientOptions = { subject };
+  }
+
+  return options;
+}
+
+/**
  * Create an authorized client from a service account JSON key file.
  * `GoogleAuth` handles JWT signing and token refresh automatically.
  */
 export async function createServiceAccountAuth(): Promise<any> {
-  const keyFile = process.env.GOOGLE_APPLICATION_CREDENTIALS!;
-  console.error(`Using service account credentials from ${keyFile}`);
+  const options = buildServiceAccountAuthOptions();
+  const subject = (options.clientOptions as { subject?: string } | undefined)?.subject;
+  console.error(
+    `Using service account credentials from ${options.keyFile}` +
+      (subject ? ` (impersonating ${subject} via domain-wide delegation)` : ''),
+  );
 
-  const auth = new GoogleAuth({
-    keyFile,
-    scopes: [...DEFAULT_SCOPES],
-  });
-
+  const auth = new GoogleAuth(options);
   const client = await auth.getClient();
   console.error('Service account authentication successful');
   return client;
--- a/src/index.ts
+++ b/src/index.ts
@@ -401,6 +401,8 @@ Environment Variables:
 
   Service Account Mode:
   GOOGLE_APPLICATION_CREDENTIALS        Path to service account JSON key file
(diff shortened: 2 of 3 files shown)


src/auth/externalAuth.ts
```diff
@@ -16,18 +16,48 @@ export function isServiceAccountMode(): boolean {
...
+  return options;
+}
+
+/**
+ * Create an authorized client from a service account JSON key file.
+ * `GoogleAuth` handles JWT signing and token refresh automatically.
+ */
+export async function createServiceAccountAuth(): Promise<any> {
+  const options = buildServiceAccountAuthOptions();
+  const subject = (options.clientOptions as { subject?: string } | undefined)?.subject;
```
piotr-agier, line 54:
```ts
const subject = (options.clientOptions as { subject?: string } | undefined)?.subject;
```

The cast reads back through the abstraction we just built, and the resulting type assertion is wider than what `buildServiceAccountAuthOptions` actually puts in there. Two cleaner options, both fine:

1. Re-read the env var directly in `createServiceAccountAuth` for the log line:
   ```ts
   const subject = process.env.GOOGLE_DRIVE_MCP_SUBJECT?.trim();
   ```
   Trivial cost, no cast, and the log path stops depending on the internal shape of `GoogleAuthOptions`.

2. Have `buildServiceAccountAuthOptions` return `{ options, subject }` so callers do not have to reconstruct it.

I would take option 1 unless you anticipate other callers needing `subject` separately.

piotr-agier (commented):
Two doc gaps that are out of the diff but worth catching before this lands.

### `CHANGELOG.md` is not updated

`CHANGELOG.md:5` has an empty `[Unreleased]` section. This PR introduces one new env var (`GOOGLE_DRIVE_MCP_SUBJECT`) and starts honoring an existing one (`GOOGLE_DRIVE_MCP_SCOPES`) in service-account mode, both user-visible. A one-liner under Unreleased so it lands in the next release notes, matching the existing entry style:

```
### Features
- **auth:** support domain-wide delegation via `GOOGLE_DRIVE_MCP_SUBJECT`; honor `GOOGLE_DRIVE_MCP_SCOPES` in service-account mode (#107)
```

If the changelog is generated by release tooling rather than written by hand, please confirm and I will withdraw this.

### `README.md` does not document `GOOGLE_DRIVE_MCP_SUBJECT`

The `--help` text in `src/index.ts` is updated, but `README.md` is not. Two specific spots:

1. `README.md:957` (`### 1. Service Account Mode`) only documents `GOOGLE_APPLICATION_CREDENTIALS`. Domain-wide delegation is a real configuration mode that operators will look up here, including the prerequisite step (admin authorizing the service account's client ID under Security > API controls > Manage Domain-wide Delegation). Suggest an example block with `GOOGLE_DRIVE_MCP_SUBJECT` set, plus a note that `GOOGLE_DRIVE_MCP_SCOPES` now applies in this mode too.
2. `README.md:1331` (the env var reference table) omits `GOOGLE_DRIVE_MCP_SUBJECT`. Add a row.

The README is the canonical place users land before running `--help`, so missing this here is the higher-impact gap.

src/index.ts
```diff
@@ -401,6 +401,8 @@ Environment Variables:
 
   Service Account Mode:
   GOOGLE_APPLICATION_CREDENTIALS        Path to service account JSON key file
+  GOOGLE_DRIVE_MCP_SUBJECT              Workspace user to impersonate via domain-wide delegation (optional)
+  GOOGLE_DRIVE_MCP_SCOPES               Comma-separated scopes to request (aliases or full URLs; defaults to all Drive/Docs/Sheets/Slides/Calendar scopes)
```
piotr-agier, line 405: `GOOGLE_DRIVE_MCP_SCOPES` is listed only under "Service Account Mode", but it is also consumed by the interactive OAuth flow (`src/auth/server.ts:9`) and by `src/tools/drive.ts:235`. Operators reading `--help` for OAuth setup will not find it here.

Two reasonable fixes:

1. Promote it to a "Common Configuration" block above the mode-specific sections, alongside `GOOGLE_DRIVE_MCP_TOKEN_PATH` and `GOOGLE_DRIVE_MCP_AUTH_PORT`. That matches its actual scope.
2. Duplicate the line under both "Service Account Mode" and "External OAuth Token Mode".

Pre-existing gap on the OAuth side, but this PR is the natural place to close it since you are touching the help text already. Option 1 is the cleaner outcome.

PR #108, opened 2026-04-22
feat(docs): add createDocFromHTML tool for rich-text Google Doc creation

stefanthorpe:
## Summary

Adds a new tool, `createDocFromHTML`, that creates a fully-styled Google Doc from an HTML string in a **single** Drive API call. The Drive API's MIME conversion maps HTML tags to native Doc styles — `<h1>` → HEADING_1, `<p>` → NORMAL_TEXT, `<b>`/`<i>` for inline styles, `<ul>`/`<ol>` for lists, `<table>` for tables, inline CSS for fonts/colors/sizes.

## Why — efficiency

Building a non-trivial Google Doc with the existing tools means:

1. `createGoogleDoc` (1 call)
2. `insertText` for each section's content (N calls)
3. `formatGoogleDocParagraph` with `namedStyleType` for every paragraph — headings AND body, because body paragraphs need an explicit `NORMAL_TEXT` to avoid inheriting the previous heading's style (M calls)
4. Inline formatting per run where needed (extra calls)

A ten-section branded doc easily hits 30–50 API calls, each round-tripping through the Docs API, each representing a step where the agent can make a mistake. `createDocFromHTML` collapses all of that into a single upload: compose the whole document as HTML once, let the Drive API's built-in converter do the layout. For our team's workflow (branded memos, PRDs, meeting briefs) this reduced average doc-creation wall-clock from tens of seconds to a single request, and dropped the failure rate to near zero.

As a side effect, it also sidesteps the paragraph-style-inheritance behaviour that the step-by-step path has to work around.

## What's in the patch

- `src/tools/docs.ts` — new tool case + definition. ~80 LOC.
- `test/integration/docs-from-html.test.ts` — happy path + duplicate-name guard + folder resolution.
- `test/schema/tool-registry.test.ts` — bumps the expected tool count and adds `createDocFromHTML` to the registry assertion.

No new dependencies. No changes to `package.json` / engines / existing tools.

## Test plan

- [x] `npm run build` — passes
- [x] `npm test` — 283/283 pass (was 281 before this patch, +2 for the new integration tests)
- [ ] Maintainer sanity check: create a doc with mixed headings, body paragraphs, bold/italic, a bullet list, a table; inspect in Google Docs and confirm native styles apply

## Provenance

Originally authored by a coworker at Relevant Search against their uncommitted tree. We've been using this tool in production for our team's branded-doc workflow and wanted to offer it back upstream. Happy to split the commit, rename, adjust the description, or drop the duplicate-name guard if you'd prefer a narrower tool — let us know.

### DIFF
--- a/src/tools/docs.ts
+++ b/src/tools/docs.ts
@@ -1,3 +1,4 @@
+import { Readable } from 'stream';
 import { z } from 'zod';
 import JSZip from 'jszip';
 import type { ToolDefinition, ToolContext, ToolResult } from '../types.js';
@@ -709,6 +710,12 @@ const CreateGoogleDocSchema = z.object({
   parentFolderId: z.string().optional()
 });
 
+const CreateDocFromHTMLSchema = z.object({
+  html: z.string().min(1, "HTML content is required"),
+  name: z.string().min(1, "Document name is required"),
+  parentFolderId: z.string().optional(),
+});
+
 const UpdateGoogleDocSchema = z.object({
   documentId: z.string().min(1, "Document ID is required"),
   content: z.string(),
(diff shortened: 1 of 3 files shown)


piotr-agier:
Could you rebase to latest master @stefanthorpe ?

PR #126, opened 2026-05-16
fix(docs): honor tabId in table/format/footnote/smartchip editing handlers (#114)

### LINKED ISSUE #114: `insertTable` / `editTableCell` silently ignore `tabId` and always operate on the default tab
### Summary

When working with a multi-tab Google Doc, `insertTable` and `editTableCell` accept a `tabId` parameter without erroring, but **silently ignore it** and place the table into the document's default (first) tab. The same misbehaviour likely affects `insertTableRow`, `addNamedRange`, `applyTextStyle`, `applyParagraphStyle`, and other body-edit tools — but I've only verified `insertTable` and `editTableCell` first-hand.

By contrast, `insertText`, `deleteRange`, and `updateGoogleDoc` correctly honour `tabId` and operate on the specified tab.

### Reproduction

1. Create a multi-tab Google Doc with at least two tabs (call them `t.0` for the default and `t.X` for a second tab).
2. Call `insertTable` with `tabId: "t.X"`, `rows: 2`, `columns: 2`, `index: 1`.
3. The tool returns success: `Successfully inserted 2x2 table at index 1`.
4. Inspect the document via `readGoogleDoc` — the new table is in `t.0`, not `t.X`.

### Expected behaviour

The table is inserted into the tab specified by `tabId`. Mirror the working behaviour of `insertText` / `deleteRange`.

### Why this is a server-side bug, not an API limitation

The underlying Google Docs API fully supports per-tab targeting for table operations. Per the official [Work with tabs](https://developers.google.com/workspace/docs/api/how-tos/tabs) and [Working with tables](https://developers.google.com/workspace/docs/api/how-tos/tables) guides, `InsertTableRequest`, `InsertTableRowRequest`, `UpdateTableCellStyleRequest`, etc. all accept a `Location.tabId` field. The Google Docs guide is explicit: *"Each Request includes a way to specify the tabs to apply the update to."*

The fix appears to be passing the `tabId` from the tool parameter through into the `Location` object on the request body — likely 2–3 lines per affected handler.

### Impact

This silently corrupts multi-tab documents: tables intended for a non-default tab end up in the default tab, often pushing existing content sideways or breaking the document structure. Because the tool reports success, the caller has no signal that anything went wrong without manually reading the document back.

### Suggested fix scope

In priority order:

1. `insertTable` — pass `tabId` through to `Location`.
2. `editTableCell` — same.
3. `insertTableRow` / any row/column manipulation tools.
4. `applyTextStyle`, `applyParagraphStyle`, `formatGoogleDocText`, `formatGoogleDocParagraph` — accept and pass through `tabId` for both `textToFind` and explicit-index modes.
5. `findAndReplaceInDoc` — already has a `tabId` parameter for scoped replacement; verify it's actually wired through (haven't tested).

### Environment

- Server version: `@piotr-agier/google-drive-mcp@2.2.0`
- Client: Claude Code 2.1.118 on macOS
- Reproduced on a doc with 4 tabs

Happy to send a small PR if useful.

---

## Optional addendum (also worth flagging in the same issue or separately)

The README and the Claude-Code-side tool schema **don't document `tabId` as an accepted parameter on most tools**, even though several tools (e.g. `insertText`, `deleteRange`, `updateGoogleDoc`) do honour it. This makes the API surface confusing — users either don't know `tabId` is available, or assume it works everywhere because it's accepted without erroring. Adding a single line per tool description like *"`tabId` (optional): scope this operation to a specific tab"* would help a lot.

piotr-agier:
## Summary

Fixes #114 — a silent data-corruption bug. Several Google Docs editing/formatting tools accepted (or should accept) a `tabId` parameter but ignored it, operating on the document's **default tab** while reporting success. In multi-tab documents this inserted tables/text/styles/footnotes into the wrong tab, often shifting or breaking existing content, with no error signal to the caller.

A full audit found the affected family is exactly **7 handlers** in `src/tools/docs.ts`. The already-correct `insertText` / `deleteRange` / `updateGoogleDoc` / `findAndReplaceInDoc` tools were used as the reference patterns.

## What changed

- **Schemas + tool definitions:** added `tabId` to the Zod schema and the MCP tool-definition `properties` (incl. the `formatGoogleDoc*` aliases) for: `insertTable`, `editTableCell`, `insertSmartChip`, `createFootnote`, `applyTextStyle`/`formatGoogleDocText`, `applyParagraphStyle`/`formatGoogleDocParagraph`, `createParagraphBullets`. Without the tool-def property an MCP client can't pass `tabId` at all.
- **Tier 1 (mechanical):** `insertTable`, `insertSmartChip`, `createFootnote` thread `tabId` into the `location`/`endOfSegmentLocation`. `createFootnote` also threads it into the footnote-body `insertText` (kept alongside `segmentId`, which are complementary — `tabId` selects the tab, `segmentId` the footnote segment within it).
- **Tier 2 (format tools):** new shared `getTabBodyContent` helper (omit `fields`, `includeTabsContent: true`, `findTabById`) makes `findTextRange`/`getParagraphRange` tab-aware so `textToFind`/`indexWithinParagraph` resolve against the right tab; explicit-index mode threads `tabId` straight into the request range via `buildUpdate*StyleRequest`.
- **Tier 3 (`editTableCell`):** previously fetched only the default body (`fields: 'body(content)'`) and was structurally tab-blind. Now resolves the target tab's body, locates the table there, and threads `tabId` into all four mutation requests.
- Unresolvable tabs return the standard `Tab with ID "..." not found. Use listDocumentTabs to see available tabs.` error.
- **Default (no-`tabId`) behavior is preserved byte-for-byte** — every default code path is unchanged; `tabId` keys are only ever added when a `tabId` is supplied.
- Success messages now append ` in tab X` when tab-scoped (matching `insertText`/`deleteRange`), giving the caller a visible signal.

## Testing

- `npm run build` ✅ (typecheck clean)
- `npm run lint` ✅
- `npm test` ✅ — **314 tests, 0 failures**
- Added integration coverage in `test/integration/docs.test.ts` for each handler: default (no `tabId`, request carries no `tabId`), tab-scoped (`tabId` threaded + message mentions it + `includeTabsContent` used where applicable), and unknown-`tabId` (standard not-found error, no `batchUpdate` issued). `createFootnote` also asserts the body insert carries **both** `segmentId` and `tabId`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ All notable changes to this project will be documented in this file.
 
 ### Bug Fixes
 
-- **docs:** honor `tabId` in `insertTable`, `editTableCell`, `insertSmartChip`, `createFootnote`, `applyTextStyle`/`formatGoogleDocText`, `applyParagraphStyle`/`formatGoogleDocParagraph`, and `createParagraphBullets` — these previously ignored `tabId` and silently edited the default tab of multi-tab documents while reporting success ([#114](https://github.com/piotr-agier/google-drive-mcp/issues/114), [#PRNUM](https://github.com/piotr-agier/google-drive-mcp/pull/PRNUM))
+- **docs:** honor `tabId` in `insertTable`, `editTableCell`, `insertSmartChip`, `createFootnote`, `applyTextStyle`/`formatGoogleDocText`, `applyParagraphStyle`/`formatGoogleDocParagraph`, and `createParagraphBullets` — these previously ignored `tabId` and silently edited the default tab of multi-tab documents while reporting success ([#114](https://github.com/piotr-agier/google-drive-mcp/issues/114), [#126](https://github.com/piotr-agier/google-drive-mcp/pull/126))
 - **auth:** use loopback IP `127.0.0.1` instead of `localhost` for the OAuth callback redirect URI, matching the IPv4-only callback-server bind so the redirect resolves to the bound address on dual-stack hosts ([#124](https://github.com/piotr-agier/google-drive-mcp/pull/124)). Desktop-app OAuth clients (the recommended type) are unaffected; "Web application" clients that registered a `http://localhost:<port>` redirect must re-register it as `http://127.0.0.1:<port>` or auth fails with `redirect_uri_mismatch` — see README Troubleshooting
 
 ## [2.2.0](https://github.com/piotr-agier/google-drive-mcp/compare/v2.1.0...v2.2.0) (2026-04-20)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ All notable changes to this project will be documented in this file.
 
 ### Bug Fixes
 
+- **docs:** honor `tabId` in `insertTable`, `editTableCell`, `insertSmartChip`, `createFootnote`, `applyTextStyle`/`formatGoogleDocText`, `applyParagraphStyle`/`formatGoogleDocParagraph`, and `createParagraphBullets` — these previously ignored `tabId` and silently edited the default tab of multi-tab documents while reporting success ([#114](https://github.com/piotr-agier/google-drive-mcp/issues/114), [#PRNUM](https://github.com/piotr-agier/google-drive-mcp/pull/PRNUM))
 - **auth:** use loopback IP `127.0.0.1` instead of `localhost` for the OAuth callback redirect URI, matching the IPv4-only callback-server bind so the redirect resolves to the bound address on dual-stack hosts ([#124](https://github.com/piotr-agier/google-drive-mcp/pull/124)). Desktop-app OAuth clients (the recommended type) are unaffected; "Web application" clients that registered a `http://localhost:<port>` redirect must re-register it as `http://127.0.0.1:<port>` or auth fails with `redirect_uri_mismatch` — see README Troubleshooting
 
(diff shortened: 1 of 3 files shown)


piotr-agier:
Pushed `78493cf` addressing the actionable review-triage notes (no behavior change):

- **Note 1 — idiom consolidation:** the three "add `tabId` iff present" spellings (conditional-assign, inline spread, local closure) are now a single module-level `withTab<T>` helper used by every tab-threading site introduced here. Pre-existing `insertText`/`deleteRange`/`updateGoogleDoc` left as-is (out of scope for this fix).
- **Note 2 — error-UX asymmetry:** added a code comment at `insertTable`/`insertSmartChip`/`createFootnote` documenting that `tabId` is passed through *unvalidated by design* so the pure-`batchUpdate` happy path stays GET-free. Deliberately did **not** add validation (would add a GET to the happy path for a marginal error-message gain).
- **Note 4 — default-path coverage:** added no-`tabId` tests for `applyTextStyle`/`applyParagraphStyle`/`createParagraphBullets` asserting the narrow-field GET (never `includeTabsContent`) is still used and no `tabId` leaks into the range — directly guarding the "byte-for-byte preserved" `findTextRange`/`getParagraphRange` path.

**Note 3** (two full-document GETs on the `applyParagraphStyle` + `textToFind` + `tabId` path) is intentionally **deferred** — the only clean fix threads pre-resolved content through the two most complex shared helpers across ~5 call sites; that's regression risk disproportionate to a bounded perf gain inside a focused bug-fix PR. Worth a separate follow-up issue.

Typecheck, lint, and the full suite (317 tests) are green.

PR #129, opened 2026-05-16
feat(calendar): support event attachments (read + write)

### LINKED ISSUE #110: Feature Request: Calendar Event attachments support
Google Calendar events can have file attachments, which the [API](https://developers.google.com/workspace/calendar/api/v3/reference/events) returns as the property `attachments[]`. It would be useful if this property was part of the `getCalendarEvent` tool response.

piotr-agier:
Closes #110.

## Summary

Adds Google Calendar event **attachment** support, both reading and writing.

### Read (the issue ask)
- `getCalendarEvent` and `getCalendarEvents` now surface event `attachments` (title + URL) in their response. No new scopes — the Calendar API already returns the field.

### Write (extra, agreed in discussion)
- `createCalendarEvent` / `updateCalendarEvent` accept an `attachments` array: `fileUrl` (required), `title`/`mimeType` (optional), **max 25** (Calendar API limit).
- The `supportsAttachments` API flag is set on insert/update so attachments are actually persisted (the API ignores them otherwise).
- For Drive files, pass the file's share URL as `fileUrl`; non-Drive URLs are also accepted. `fileId`/`mimeType` are output-only and derived by the API.

### Bug fix (found along the way)
`buildCalendarEventUpdate` never carried `attachments` through, so **every** `updateCalendarEvent` call silently wiped an event's existing attachments. It now preserves them when `attachments` is omitted and replaces them when supplied.

## API basis
Verified against the [Calendar API events reference](https://developers.google.com/workspace/calendar/api/v3/reference/events): 25-attachment cap, `fileUrl` required + writable, `fileId`/`mimeType` output-only, `supportsAttachments` required to persist attachment changes.

## Tests
- 5 integration tests (read surfaces attachments; create forwards `attachments` + `supportsAttachments`; missing-`fileUrl` rejected; update preserves existing; update replaces when provided)
- 2 unit tests for the `buildCalendarEventUpdate` preserve/replace behavior
- Full suite green on a clean `master` base: **97/97 unit, 324/324 integration**, typecheck clean, tool count unchanged (107)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ All notable changes to this project will be documented in this file.
 
 ### Features
 
+- **calendar:** surface event `attachments` in `getCalendarEvent`/`getCalendarEvents` responses, and accept an `attachments` array (max 25) in `createCalendarEvent`/`updateCalendarEvent` (sets the `supportsAttachments` API flag). `updateCalendarEvent` now also preserves an event's existing attachments instead of silently dropping them when `attachments` is not supplied ([#110](https://github.com/piotr-agier/google-drive-mcp/issues/110))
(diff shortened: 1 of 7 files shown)


benchun:
Thank you!

Issue #109, opened 2026-04-22
Design proposal: optional OAuth 2.1 authorization server for team / multi-user deployments


stefanthorpe:
Opening this as a design proposal rather than a PR, because the changes are large and opinionated. Happy to break it into pieces if any of it sounds interesting to land upstream.

## Context

Over at RS we deploy this server on Cloud Run behind \`https://drive-mcp.relevantsearch.com\` and expose it to our team via the claude.ai custom-connector flow. The current server supports single-user auth (local OAuth with tokens on disk, or stdio mode with env-var credentials) — both of those break down when the server is a shared service and every team member needs to act as themselves against their own Drive.

Our fork adds an optional \"team mode\" where the HTTP transport becomes an MCP-spec OAuth 2.1 authorization server per [RFC 9728](https://datatracker.ietf.org/doc/html/rfc9728) / the MCP auth spec. Each user hits a sign-in flow, the server persists their refresh token, and every subsequent tool call acts as them. We'd like to offer the pattern back upstream if there's appetite.

## What changes

1. **Custom \`OAuthServerProvider\`** implementing all six methods (\`clientsStore\`, \`authorize\`, \`challengeForAuthorizationCode\`, \`exchangeAuthorizationCode\`, \`exchangeRefreshToken\`, \`verifyAccessToken\`). Rejected \`ProxyOAuthServerProvider\` — Google doesn't support Dynamic Client Registration, so the proxy variant can't forward claude.ai's dynamically-issued \`client_id\` upstream.

2. **Two-hop OAuth**: claude.ai ⇄ our server (AS) ⇄ Google (IdP). We stash pending authorizations keyed by Google-side \`state\` so the round-trip carries claude.ai's PKCE challenge + redirect + client_id through the Google consent flow.

3. **SDK \`mcpAuthRouter\`** mounted at the app root — gives us \`/register\`, \`/authorize\`, \`/token\`, and the \`.well-known\` discovery endpoints for free. \`requireBearerAuth\` guards every \`/mcp\` route.

4. **Per-user tool context**: each MCP request rebuilds the \`ToolContext\` with a fresh \`OAuth2Client\` seeded from that user's persisted refresh token, so every tool acts as the caller. A \`tokens\` listener persists Google's auto-refreshed access tokens back to storage.

[...]

PR #145, opened 2026-07-11
feat: team mode — OAuth 2.1 authorization server for multi-user deployments (#109)

piotr-agier:
Implements the team-mode proposal from #109: an opt-in mode where the HTTP transport becomes an MCP-spec OAuth 2.1 authorization server, each team member signs in with their own Google account, and every tool call runs as the caller. Designed for exposing the server to a team through claude.ai custom connectors.

## Architecture

Two-hop OAuth: MCP client (Dynamic Client Registration + PKCE S256) ⇄ this server (AS) ⇄ Google (IdP). A full `OAuthServerProvider` rather than `ProxyOAuthServerProvider`, because claude.ai requires DCR and Google doesn't support it. The SDK's `mcpAuthRouter` provides `/authorize`, `/token`, `/register`, `/revoke` and the RFC 8414/9728 discovery metadata; `requireBearerAuth` guards the three `/mcp` routes with the `resource_metadata` pointer that drives client-side OAuth discovery.

Layered for review (one commit per layer):

1. `fix(auth)` — pre-existing log leak: raw gaxios errors embed the token-refresh POST body (refresh token + client secret) in `err.config`; JSON.parse errors echo token-file fragments. All auth-layer error logs now extract known-safe fields only.
2. `feat(team)` — pluggable `TeamStore` (in-memory + `team-store.json` file backend reusing the AccountStore write discipline: serialized queue, merge-from-disk, temp+rename, 0600). Atomic single-use consumption, hard caps on every collection, per-sub token eviction.
3. `feat(team)` — `TeamOAuthProvider` + two-hop flow + Google callback behind `--team` / `MCP_TEAM_MODE`.
4. `feat(team)` — bearer-guarded per-user dispatch: identity comes exclusively from the bearer's `sub` (`extra.authInfo`), per-sub client caches in `TeamClientFactory`, session↔user binding.
5. `docs(team)` — README Team Mode section, security notes, CHANGELOG.

## Deliberate deviations from the #109 proposal

- **Opaque server-stored tokens instead of HS256 JWTs.** A store is required either way (refresh tokens, DCR clients, per-user Google grants), and `verifyAccessToken` must hit it regardless to load the caller's identity — so JWT statelessness buys nothing in a single-process deployment, while opaque tokens remove the `JWT_SECRET` management surface and give exact revocation for free (needed for the `invalid_grant` recovery flow). Tokens are 256-bit random, SHA-256-hashed at rest.
- **Filesystem instead of Firestore as the canonical persistent store.** The `TeamStore` interface is the extension point; a Firestore adapter can live out of tree.

## Security properties (each pinned by a test)

- SDK verifies PKCE S256 locally; the provider enforces everything the SDK leaves open: code↔client binding, `redirect_uri`/`resource` re-validation at exchange, single-use codes (atomic consume; burned after 3 challenge lookups to bound verifier brute-force).
- Refresh rotation with a 60s grace window; post-grace reuse revokes the whole token family (OAuth 2.1 theft response). Superseded tokens are kept as tombstones so reuse stays detectable.
- `prompt=consent` on every Google authorization — the confused-deputy mitigation for a static upstream client behind open DCR (silent re-consent would let an attacker-registered client mint tokens for anyone who clicks a link). Optional `MCP_TEAM_ALLOWED_REDIRECT_URIS` narrows DCR further.
- Hosted-domain allowlist enforced on the Google-asserted `hd` claim (never the email suffix), failing closed for consumer accounts, revoking the fresh grant on rejection.
- Sessions bind to the initializing user's `sub`; a foreign bearer gets the same 400 as an unknown session, so leaked session ids are useless and unprobeable.
- The `account` parameter, `manage_accounts`, and the resources capability (which reads the shared default account) are disabled in team mode.
- `AuthInfo.expiresAt` is epoch seconds (a ms value would make tokens immortal against `requireBearerAuth`'s check) — structurally asserted.
- Callback failures before the pending record is validated render a static 400 page, never a redirect (open-redirect prevention); callback inputs are never reflected or logged.
- Host-header validation derived from the issuer (the SDK applies none for non-localhost binds); `MCP_TRUST_PROXY` keeps per-IP rate limits meaningful behind Cloud Run/nginx.
- A revoked Google grant self-heals: the in-flight call returns a reconnect hint, the user's MCP tokens are dropped so the next request 401s the connector into a fresh OAuth flow.

## Compatibility

Default stdio/HTTP behavior is byte-for-byte unchanged: team code is only reachable behind the flag, never calls `buildAuthSystem` (no interactive boot consent can fire), and never touches `tokens.json`. Team mode is mutually exclusive with service-account/external-token modes (hard startup error). Requires a "Web application" Google OAuth client; `loadWebCredentials` rejects desktop-type keys with the exact redirect URI to register. Single-process deployment is a documented constraint (`--max-instances=1`).

## Testing

63 new tests: store contract suite against both backends (incl. concurrent double-consume, crash-mid-write, 0600, corrupt-file quarantine), provider unit tests, full two-hop flow over real HTTP with a faked IdP, dispatch integration (401 matrix, two users routing to their own clients, session hijack, scope gates, invalid_grant self-heal), CLI validation against the built binary. Full suite: 698/698.

Also hardened three pre-existing test-suite flakes that failed under full parallel load (fixed 3s spawnSync timeout in cli-args, 50ms idle-timer margins, 2s waitFor budget).

Manual end-to-end verification against real Google (web client + https tunnel + claude.ai connector, two accounts) is parked in `backlog/todo/` to run before tagging a release.

## Code-review fixes (post-review)

A multi-agent review of this branch surfaced **15 confirmed findings**, all now fixed on the branch (commit `fix(team): resolve code-review findings on the OAuth 2.1 server`). Each fix reuses existing helpers and is pinned by a regression test.

**Deployment-breaking**
- DCR client cap now LRU-evicts the idle registration instead of throwing — an unauthenticated `/register` flood could otherwise permanently lock out onboarding. Recency is a monotonic tick persisted in a new optional `clientMeta` file section; token issuance keeps active connectors ahead of idle ones.
- `fileStore.init` quarantines only genuine content corruption (bad JSON / wrong shape) and **rethrows on IO/permission errors** (EACCES, EIO) instead of renaming a possibly-intact store aside — a transient glitch could otherwise wipe every member's Google grant.
- `redirect_uri` and `resource` are compared only when the token request sends them (OAuth 2.1 drops `redirect_uri`; RFC 8707 treats `resource` as per-request), inheriting the stored value otherwise — a spec-compliant client is no longer dead-ended after its single-use code is consumed.
- Refresh-scope narrowing binds the issued access token only; the rotated refresh token keeps the full grant (RFC 6749 §6), so a one-time narrowing no longer permanently shrinks the grant.

**Security / correctness**
- Dispatch enforces the bearer token's **own** scopes (`extra.authInfo.scopes`), not just the Google grant, so a narrowed or leaked token can't reach every tool the user consented to.
- A domain-allowlist rejection revokes the fresh grant only for a brand-new sub; an existing member's stored grant is left intact, so a transient `hd` omission no longer destroys their working refresh token at Google.
- A new atomic `updateUser(sub, mutator)` store primitive backs the client factory's revoked-grant and token-refresh paths, so a concurrent re-authorization is not clobbered by a stale snapshot.
- Credential-load and startup errors are routed through `describeErrorForLog`, closing a residual leak where a malformed `gcp-oauth.keys.json` could echo the `client_secret` into logs.

**Operational**
- `MCP_HTTP_ALLOWED_HOSTS` entries are lowercased for the SDK's case-sensitive host check; a path-bearing issuer URL is rejected at boot (it would be silently dropped from the derived Google callback).
- The team+HTTP transport guard applies only on server-start, so `MCP_TEAM_MODE` no longer aborts `version`/`help`/`auth`.
- `authGetStatus`/`authListScopes` are hidden in team mode and rejected if called (they report local `tokens.json` state).
- A store `flush()` drains the file-store write queue on shutdown, so a just-completed sign-in or token rotation is not lost.
- Superseded-token tombstones are bounded to a 24h reuse-detection horizon plus a per-sub cap, instead of being retained for the full 30-day grant TTL.

Two behaviours were confirmed with the maintainer before implementing: **LRU eviction** for the client cap, and **preserve-existing-member** on domain rejection.

Full suite now: **724/724** (was 698).


piotr-agier:
Superseded by #148, which contains the identical changes rebased linearly onto current `master`. This PR's header was stuck showing a stale "merge conflicts" banner even though the GitHub API reported it mergeable/clean; recreating on a fresh branch clears it. Closing in favor of #148.

PR #157, opened 2026-07-17
feat(distribution): publish official MCP Registry metadata

piotr-agier:
## Summary

- add validated `server.json` metadata for `io.github.piotr-agier/google-drive-mcp`
- align npm metadata and version `2.5.1` with the immutable Registry release contract
- validate metadata in CI and publish it after npm via GitHub OIDC
- add idempotent post-publication verification, contract tests, and maintainer documentation

## Verification

- `npm run lint`
- `npm run build`
- `npm run registry:validate`
- `npm test` — 774 tests passing
- packed npm manifest contains the required `mcpName`
- Registry helper missing/existing-record paths tested against a local mock

## Release behavior

Publishing GitHub release `v2.5.1` first publishes the npm artifact, then publishes and verifies the matching official MCP Registry record.


piotr-agier:
## Addressed review findings (commit c33b8aa)

Fixed the **confirmed** findings from the review:

| Finding | Fix |
|---|---|
| Order-sensitive `comparePublishedRecord` fails post-publish verify on a correct release | Order-insensitive **subset match** — ignores registry key reordering and injected fields; only the metadata we own is asserted |
| CI coupled to live registry on every push/PR | New script exit-code taxonomy (`1`=invalid, `2`=not-found, `3`=unreachable); CI **soft-passes on registry outage** (exit 3 → warning) but still blocks on genuinely invalid metadata (exit 1) |
| npm release depends on registry reachability | Removed `registry:validate` from the `publish` job — `npm publish` no longer sits behind a live-registry call |
| Duplicated validate + version-check across jobs | Release-time validate now runs once (in `publish-registry`); dropped the duplicated version-vs-tag check |
| Dead 4xx fast-fail branch (client errors retried) | Subsumed by the retry unification below — `withRetry` never retries 4xx |
| Hand-rolled retry duplicates `src/utils/retry.ts` | Extracted a shared plain-JS core (`src/utils/retry.core.js`) consumed by both the bundled app (via `retry.ts`) and the standalone registry script — no build/`node_modules` needed |
| No test guarding `description` drift | Added a `package.json` ↔ `server.json` description-parity assertion |

**Verification:** `typecheck` / `lint` / `build` clean, **775/775 tests** pass (incl. the existing `retry.test.ts` behavioral suite). The exit-code contract (`0/1/2/3`), 4xx fast-fail, subset-match, and the CI soft-fail wrapper were each exercised against a local registry stub.

_Note:_ adopting `withRetry` narrows retry policy so `500/502` are no longer retried (only `429/503/504` + timeout/network) — they still classify as *unreachable* (soft in CI). Deliberate, matching the app's existing stance for these idempotent registry calls.

The **plausible** findings (re-run recovery on verify mismatch, verify indexing window, npm-propagation ownership race, search pagination, and the additive-vs-PATCH semver level) were captured as separate follow-up items for further verification rather than changed here.

Issue #137, opened 2026-06-24
Shared Drives invisible & search returns 0 results (incl. My Drive) on 2.2.0 despite full auth/drive scope


jasonlevin-cmyk:
## Summary

On `2.2.0` (the current latest), the server cannot see any Shared Drive content **and** `search` returns zero results even for My Drive, despite the OAuth token holding the full `https://www.googleapis.com/auth/drive` scope and the same Google account being able to reach the files fine via the Drive REST API directly. There's no auth error — operations just come back empty / "file not found." So this looks broader than the usual "missing `supportsAllDrives`" Shared-Drive gap.

## Environment

- Package: `@piotr-agier/google-drive-mcp@2.2.0` (latest release)
- Launched via: `npx -y @piotr-agier/google-drive-mcp@2.2.0` (stdio), `cmd /c` wrapper
- Client: Claude Code (MCP stdio client)
- OS: Windows 11
- Node: v24.14.0, npm: 11.9.0
- Auth: desktop OAuth flow, `GOOGLE_DRIVE_OAUTH_CREDENTIALS` pointing at an OAuth client credentials JSON; token persisted to `tokens.json`
- Account type: Google **Workspace** account (Shared Drives are in use org-wide)

## Granted scopes (from the server's own `tokens.json`)

The persisted token's `scope` field contains the full set:

```
https://www.googleapis.com/auth/drive
https://www.googleapis.com/auth/drive.readonly
https://www.googleapis.com/auth/drive.file
https://www.googleapis.com/auth/documents
https://www.googleapis.com/auth/spreadsheets
https://www.googleapis.com/auth/presentations
https://www.googleapis.com/auth/calendar
https://www.googleapis.com/auth/calendar.events
```

So this does **not** appear to be a too-narrow (`drive.file`-only) scope problem — the broad `auth/drive` scope is present, and `authGetStatus` shows a valid, auto-refreshing token.

## Symptoms (actual tool outputs)

- `search` with a common term (e.g. `test`) → `Found 0 files:`
- `search` with `rawQuery: true` and `name contains 'test'` → `Found 0 files:`
- `listSharedDrives` → `No shared drives found.`
- `authTestFileAccess <a-real-shared-drive-file-id>` → `File not found: <id>.`

The account genuinely owns/has access to many My Drive files and is a member of several Shared Drives (all visible in the Drive web UI).

## What I verified

[...]

PR #176, opened 2026-08-21
fix(deps): upgrade Google API clients to gaxios 7 for Node 24 compatibility

### LINKED ISSUE #166: OAuth token refresh fails on Node 24 with "Premature close" (gaxios 6 / node-fetch 2)
### Problem Description
  
    When running `@piotr-agier/google-drive-mcp` under **Node.js 24** (e.g., v24.17.0), authenticating or refreshing OAuth tokens fails with the following error:
  
    ```text
    Authentication Failed
    Invalid response body while trying to fetch https://oauth2.googleapis.com/token: Premature close
  
  ### Root Cause
  
  The package specifies google-auth-library@^9.15.0 and googleapis@^144, which resolve to gaxios@6.7.1 -> node-fetch@2.7.0. Under Node 24 runtime, node-fetch@2 encounters a Premature     
  close error when POSTing to Google's OAuth token endpoint (https://oauth2.googleapis.com/token).
  
  Testing isolated client behavior against the same token endpoint:
  
  • curl -X POST → HTTP 400 (reaches endpoint fine)
  • Node native fetch → HTTP 400 (reaches endpoint fine)
  • gaxios 6.7.1 request() → ❌ Premature close
  • gaxios 7.1.x request() → ✅ HTTP 400
  
  ### Suggested Fix
  
  Upgrading google-auth-library and googleapis dependencies to versions utilizing gaxios 7 (which uses native fetch rather than node-fetch@2) resolves the issue:
  
  • google-auth-library: ^10.2.0
  • googleapis: ^173.0.0
  
  Note: Overriding gaxios alone to 7 without upgrading googleapis breaks googleapis@144, which imports the legacy gaxios/build/src/common subpath removed in gaxios 7.

piotr-agier:
## Summary

Fixes #166 — on Node 24, OAuth token refresh failed with `Invalid response body while trying to fetch https://oauth2.googleapis.com/token: Premature close`. The cause is `gaxios@6` → `node-fetch@2`, which misbehaves against Google's token endpoint under Node 24. gaxios 7 ships `node-fetch@3` instead and is unaffected.

Verified live on Node 24 (see the verification comment below): the transport now returns a structured HTTP response where gaxios 6 gave "Premature close", and a forced token refresh through the server succeeds.

### Dependency changes
- Removed `@google-cloud/local-auth` — unused (zero imports), and the only thing pinning `google-auth-library@^9`.
- `google-auth-library@^10.2.0`, `googleapis@^173.0.0` (resolve to 10.9.1 / 173.0.0, gaxios 7).
- Added an npm `overrides` entry for `google-auth-library`: `googleapis-common@8` pins an **exact** version, which otherwise nests a duplicate copy and breaks `tsc` with mismatched `OAuth2Client` types.
- `node-fetch@2` is gone from the tree entirely (the remaining `node-fetch@3` is what gaxios 7 itself uses).
- Added `"engines": { "node": ">=18" }`, matching the upstream floor.

### Two regressions this upgrade introduced, found by live testing and fixed here

**1. Malformed service-account key files were silently accepted** (`97a676a`). In google-auth-library v10, `new GoogleAuth({keyFile}).getClient()` *resolves* with a credential-less JWT client when the file is unparseable or missing fields — only an absent file still throws. The server therefore logged `Service account authentication successful` while holding no credentials, and every later call failed with a misleading `Method doesn't allow unregistered callers`. v9 failed immediately. `validateCredentialsFile()` now runs before `GoogleAuth` is constructed, naming the file and the missing fields and never echoing key material. `GOOGLE_APPLICATION_CREDENTIALS` is the standard ADC variable, so non-service-account types (`authorized_user`, `external_account`, …) are deliberately still accepted; only `service_account` keys are checked field-by-field, and a missing file still surfaces as `ENOENT`.

**2. Every image was mislabelled `application/octet-stream`** (`3504826`). gaxios 7 exposes `response.headers` as a `Headers` instance where v6 returned a plain object, so `getGoogleDocImage`'s bracket-access content-type lookup always fell through to its default — including in the MCP `type: "image"` block clients use to render. `getResponseHeader()` accepts both shapes case-insensitively. This was the only gaxios response-header read in the codebase. Verified live: an inline PNG now reports `image/png`.

### Defensive adaptations (not fixes for observed breakage)

gaxios 7's README describes native-fetch semantics, but `Gaxios.#getFetch()` resolves to node-fetch v3 whenever `window` is undefined — so on Node the old shapes largely persist. Measured rather than assumed:

- **`responseType: 'stream'` still yields a Node `Readable`** (a `PassThrough`), confirmed against the live Drive API. `toNodeReadable()` normalizes web `ReadableStream`s too, so the `.on()` reader in `downloadTextContent` and the `restoreRevision` download→upload round trip stay correct if the default changes or a caller supplies its own `fetchImplementation`.
- **Transport errors still carry a top-level string `code`** (`ECONNREFUSED` observed), so the previous flat check kept working. The retry policy now walks the cause chain anyway, which also covers wrappers that only set a code on a cause, and accepts undici's `UND_ERR_CONNECT_TIMEOUT` / `UND_ERR_SOCKET`.
- `docs.ts` 404/403 mapping prefers the v7 top-level numeric `error.status`; v7 sets a numeric `code` only for AIP-193-shaped errors.
- No change needed, verified: `isInvalidGrant` (`response.data.error` survives), `describeErrorForLog` (allowlist reads only preserved fields), upload bodies (Node `Readable` still accepted), the three `responseType: 'arraybuffer'` sites (`Buffer.from` accepts both), AbortSignal forwarding.

### CI
Matrix now runs Node **22 and 24** — Node 24 is where #166 reproduces and was previously untested.

### Tests (834 → 851)
The image regression survived 834 passing tests because the doubles were built in the gaxios 6 shape: `getGoogleDocImage` already asserted `mimeType === 'image/png'`, but its mock returned `headers` as a plain object. Right assertion, wrong fixture.

`test/gaxios-contract.test.ts` (`479e0e2`) now pins the real library's shapes against a local HTTP server — no network, no credentials — without importing gaxios types: `Headers`-instance headers, Node-Readable stream bodies, numeric `status` with parsed `response.data`, a string `code` in the transport-error cause chain, and `GoogleAuth`'s acceptance of a malformed key file. Each failure message names the assumption and the source file that depends on it. The `getGoogleDocImage` doubles now use `new Headers(...)` — reintroducing the bug fails them, verified — with one plain-object case kept for the back-compat branch. Plus unit tests for `validateCredentialsFile`, `getResponseHeader`, `toNodeReadable`, and the retry cause-chain walk.

## Verification

- `npm run lint`, `npm run build`, `npm test` — 851/851 pass; CI green on Node 22 and 24.
- `npm ls google-auth-library gaxios node-fetch` — single deduped `google-auth-library@10.9.1`, gaxios 7 only, no `node-fetch@2`.
- **Live on Node 24** against real Google APIs: the #166 repro at transport and server level; `readTextFile`, both `restoreRevision` branches, `downloadFile` (binary + Workspace export → valid PDF), `uploadFile` create and in-place; all four auth modes (local OAuth, external token, full external set exercising its own refresh path, service account with a throwaway SA); and an HTTP-transport round trip with session isolation and clean SIGINT/SIGTERM shutdown. No "Premature close" in any run.
- Not covered: team mode (automated tests only), domain-wide delegation (needs a Workspace domain), and the Docker rows.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [22]
+        node-version: [22, 24]
 
     steps:
       - uses: actions/checkout@v5
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,12 +9,11 @@
             "version": "2.6.0",
             "license": "MIT",
             "dependencies": {
-                "@google-cloud/local-auth": "^3.0.1",
                 "@modelcontextprotocol/sdk": "^1.24.0",
                 "@types/express": "^5.0.3",
                 "express": "^5.1.0",
-                "google-auth-library": "^9.15.0",
-                "googleapis": "^144.0.0",
+                "google-auth-library": "^10.2.0",
+                "googleapis": "^173.0.0",
                 "jszip": "^3.10.1",
                 "open": "^10.2.0",
                 "pdf-lib": "^1.17.1",
@@ -34,6 +33,9 @@
                 "eslint": "^9.39.3",
                 "shx": "^0.4.0",
                 "typescript": "^5.6.2"
+            },
+            "engines": {
+                "node": ">=18"
             }
         },
         "node_modules/@esbuild/aix-ppc64": {
@@ -523,9 +525,9 @@
             }
         },
         "node_modules/@eslint/config-array/node_modules/brace-expansion": {
-            "version": "1.1.12",
-            "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-1.1.12.tgz",
-            "integrity": "sha512-9T9UjW3r0UW5c1Q7GTwllptXwhvYmEzFhzMfZ9H7FQWt+uZePjZPjBP/W1ZEyZ1twGWom5/56TF4lPcqjnDHcg==",
+            "version": "1.1.18",
+            "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-1.1.18.tgz",
+            "integrity": "sha512-Edep/X9fGqVNmzKBVsDYIOtD+z1tuezV70LBjdCst9Tqu76lsnvRiZ6oTic1n+/BIwX6QDGAO94PN4N2SADvtw==",
             "dev": true,
             "license": "MIT",
             "dependencies": {
@@ -597,9 +599,9 @@
             }
         },
         "node_modules/@eslint/eslintrc/node_modules/brace-expansion": {
-            "version": "1.1.12",
-            "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-1.1.12.tgz",
-            "integrity": "sha512-9T9UjW3r0UW5c1Q7GTwllptXwhvYmEzFhzMfZ9H7FQWt+uZePjZPjBP/W1ZEyZ1twGWom5/56TF4lPcqjnDHcg==",
+            "version": "1.1.18",
+            "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-1.1.18.tgz",
+            "integrity": "sha512-Edep/X9fGqVNmzKBVsDYIOtD+z1tuezV70LBjdCst9Tqu76lsnvRiZ6oTic1n+/BIwX6QDGAO94PN4N2SADvtw==",
             "dev": true,
             "license": "MIT",
             "dependencies": {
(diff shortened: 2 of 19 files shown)


piotr-agier:
## Live smoke verification on Node 24 — complete

Ran against real Google APIs on this branch (Node 24.20.0, gaxios 7.3.1, googleapis 173.0.0, google-auth-library 10.9.1). No `Premature close` in stderr in any run.

**Issue #166 verified fixed at two levels**
- *Transport:* a gaxios 7 POST to `oauth2.googleapis.com/token` returns a structured HTTP 400 — the issue's own methodology, where gaxios 6 gave "Premature close". Passed on Node 24 and 22.
- *End to end:* forced an expired-token refresh through the stdio server → refresh succeeded (`lastRefreshedAt` updated), `search` returned live results, `isError:false`.

**Stream paths** (the code whose shape changed under gaxios 7)

| check | result |
|---|---|
| `readTextFile` over a web `ReadableStream` | pass — the path that broke outright pre-fix |
| `restoreRevision`, binary (`revisions.get` stream) | pass — confirmed by re-read *and* re-download |
| `restoreRevision`, Google Doc (export-link stream) | pass — formatting warning present |
| `downloadFile` binary via `pipeline` | pass |
| `downloadFile` Workspace export → PDF | pass — valid `%PDF`, 16 KB |
| `uploadFile` create + in-place revision | pass — Node `Readable` bodies |

**All auth modes**: local OAuth; external-token with a real token (correct "no refresh token" warning); full external credential set with a deliberately invalid access token, which exercised that path's *own* refresh implementation through gaxios 7; and service-account mode with a throwaway SA (JWT exchange, shared-file read/download, correct 0-result `search` with no domain-wide delegation, #137 bypass warning, ENOENT and precedence rows).

## One regression found and fixed (85b5cc6)

Service-account testing surfaced a real behavior change from the upgrade. In google-auth-library v10, `new GoogleAuth({keyFile}).getClient()` **resolves with a credential-less JWT client** when the key file is unparseable or missing its fields — only an *absent* file still throws. The server therefore logged `Service account authentication successful` while holding no credentials, and every later call failed with a misleading `Method doesn't allow unregistered callers`. v9 failed immediately with `key must be a string, a buffer or an object`.

`validateCredentialsFile()` now runs before `GoogleAuth` is constructed, naming the file and the missing fields and never echoing key material. Since `GOOGLE_APPLICATION_CREDENTIALS` is the standard ADC variable, non-service-account types (`authorized_user` from `gcloud auth application-default login`, `external_account`, …) are deliberately still accepted; only `service_account` keys are checked field-by-field, and a missing file still surfaces as `ENOENT`.

Tests: **842 passing** (8 added).

piotr-agier:
## Correction to two claims in the PR description, plus contract tests

While adding regression coverage I tested two assumptions from this PR's description that had been taken from the gaxios 7 README rather than measured. Both were wrong.

**gaxios 7 does not use native fetch on Node.** `Gaxios.#getFetch()` resolves to `(await import('node-fetch')).default` whenever `window` is undefined — so node-fetch **v3**, not the platform fetch. Consequences:

| claim in the description | measured behavior |
|---|---|
| `responseType: 'stream'` returns a web `ReadableStream`, so `downloadTextContent`'s `.on()` reader "breaks outright" | Returns a Node `PassThrough`. Verified against the live Drive API and a local server. `.on()` works; nothing was broken. |
| transport errors bury the code at `err.cause.cause.code`, so retries "silently stop working" | `GaxiosError.code` is set at the top level (`ECONNREFUSED` observed). The original flat `err.code` check would have kept working. |

So `toNodeReadable()` and the retry cause-chain walk are **defensive, not fixes for live breakage**. They are still correct and cost nothing — gaxios documents the native-fetch behavior, and callers can set `fetchImplementation` — but the description overstated them. The two *real* regressions are unaffected: the `Headers` content-type bug (confirmed live: images were labelled `application/octet-stream`) and `GoogleAuth` silently accepting a malformed key file (reproduced directly). #166 itself is unaffected — gaxios 7 fixes it because node-fetch v3 replaces the v2 that had the "Premature close" bug.

**Why this class of mistake kept surviving CI**, and what now stops it: `getGoogleDocImage` already had three tests asserting `mimeType === 'image/png'`, and they passed throughout — the doubles returned `headers` as a plain object, the gaxios 6 shape. Right assertion, wrong fixture.

`test/gaxios-contract.test.ts` now pins the real library's shapes against a local HTTP server — no network, no credentials — without importing gaxios types: `Headers`-instance headers, Node-Readable stream bodies, numeric `status` with parsed `response.data`, a string `code` in the transport-error cause chain, and `GoogleAuth`'s acceptance of a malformed key file. Each assertion names the source file that depends on it, so a future bump fails with a pointer rather than a silent behavior change. The `getGoogleDocImage` doubles now use `new Headers(...)`; reintroducing the bug fails them (verified). One plain-object case is kept for `getResponseHeader`'s back-compat branch.

851 tests passing.

PR #210, opened 2026-09-16
feat(sheets): getGoogleSheetCells — structured, addressable cell reads

salimaCodebase:
## The problem

`getGoogleSheetContent` is currently the only way to read cell contents, and it renders a range as human-readable text — one line per row, cells joined with a tab, prefixed `Row N`. Three things cannot be recovered from that rendering.

**Addresses are relative, and columns are not addressed at all.** `Row N` counts from the start of the *requested range*, not from the sheet, so reading `'Sheet1!C6:E9'` reports its first row as `Row 1` when it is really row 6. Columns are never labelled, and since the API trims trailing empty cells, tab position does not reliably carry the column either. A caller that wants to write back to a cell it just read has to reconstruct the address by counting.

**A formula and its result require two reads.** `valueRenderOption` picks one rendering for the whole call: `FORMULA` gives the formula text, `FORMATTED_VALUE` gives the computed result. Checking a formula against what it evaluates to therefore costs two round trips over the same range, whose responses then have to be re-aligned by position.

**Everything arrives as display text.** An error cell is the literal string `#DIV/0!`, indistinguishable from a cell that genuinely contains that text. Numbers, booleans and dates arrive as whatever the sheet happened to display. Notes, hyperlinks and formats are not reachable at all. And only one range can be read per call, so an audit spanning several ranges costs one request each.

These are exactly the reads an agent makes right before it writes: locate a value, check the formula behind it, update the right cell. An address that is off by the range's own offset does not fail loudly — it writes to the wrong cell.

## What this adds

`getGoogleSheetCells`, a read-only tool backed by `spreadsheets.get` with a field mask rather than `spreadsheets.values.get`:

- **Absolute A1 address on every cell**, computed from the grid offsets the response itself reports — sheet-absolute no matter where the requested range starts.
- **Formula and result together.** `userEnteredValue` and `effectiveValue` are returned by default, so a formula and its computed value — or its typed `errorValue`, with the error type and message — are visible in one read.
- **Several ranges per call**, returned separately and in request order, each tagged with its `sheetId` and sheet title.
- **Selectable fields.** `fields` names CellData fields exactly as the Sheets API names them, so `note`, `dataValidation`, `userEnteredFormat`, `effectiveFormat`, `hyperlink` and `textFormatRuns` are all reachable.
- **Optional per-sheet metadata.** `sheetMetadata` returns merges, hidden rows and columns, frozen counts, dimension groups and dimension sizes once per sheet rather than repeated per range. Merges are reported at their full extent even when they reach outside the requested window — clipping them would hide the very fact that the merge is wider than what was read.
- **Bounded responses.** Both a cell budget and a byte budget apply. On overflow the read stops at a row boundary and returns `truncated: true` plus `nextRanges` — A1 ranges covering the unread remainder that can be passed straight back in `ranges`, so paging needs no address arithmetic from the caller. A single row larger than the whole budget is still returned, so paging always makes progress instead of looping.

The address arithmetic, field-mask construction, range-to-GridData matching and truncation live in `src/tools/sheetCells.ts`, separate from the API call, so they are unit-testable without stubbing `googleapis` — the same split `batchPassthrough.ts` already uses. Tests cover those helpers directly, plus a fixture captured from a real `spreadsheets.get` response driven through the whole pipeline.

## Backward compatibility

Purely additive: the diff removes **no line** from `src/tools/sheets.ts`.

`getGoogleSheetContent` is **untouched** — same schema, same output, same behaviour — and stays the right tool for a quick human-readable dump of a single range. No existing tool, shared helper or output shape changes. The new tool is read-only and needs no scope beyond the Sheets read scopes `getGoogleSheetContent` already uses.

Tool count 121 → 122, updated in `README.md`, `docs/tools.md` and `test/schema/tool-registry.test.ts`.

## Verification

`npm ci && npm run lint && npm run build && npm run registry:validate && npm test` — all green, 1020/1020 tests passing.


src/tools/sheetCells.ts
```diff
@@ -0,0 +1,423 @@
...
+}
+
+// Grammar for a bare (sheet-less) A1 range: optional $, up to a few letters,
+// optional $, digits, optionally repeated after a colon. Used only to tell a
+// bare RANGE ("C6:E9", "A1", "C:EB") apart from a bare, unquoted SHEET NAME
+// ("Probe") that also has no '!' of its own — Google itself resolves this
+// exact ambiguity by grammar (a string valid as a range is a range on the
+// first sheet; anything else is a sheet name), so this mirrors that rule
+// rather than inventing a different one.
+const BARE_RANGE_RE = /^\$?[A-Za-z]{0,3}\$?[0-9]*(:\$?[A-Za-z]{0,3}\$?[0-9]*)?$/;
```
piotr-agier: This regex is looser than valid A1, and the gap hits common sheet names. `[A-Za-z]{0,3}\$?[0-9]*` accepts `Jan`, `Dec`, `Log`, `Tax`, `2024`, `Q1`, and `FY25` as bare ranges, but none of those is a valid A1 range, so Google resolves each as a sheet name. The tool then routes them to the first sheet and fails with `Sheet "Summary" returned fewer grids than the ranges requested from it`, which points the caller at the wrong thing. Month and year tabs are the norm in the spreadsheets this tool targets.

The rule the comment describes is right, the regex just does not implement it. A single-part form needs both letters and digits (a cell). A two-part form needs each side to be a column, a row, or a cell. Something like:

```ts
const COL = '\\$?[A-Za-z]{1,3}';
const ROW = '\\$?[0-9]+';
const CELL = `${COL}${ROW}`;
const SIDE = `(?:${CELL}|${COL}|${ROW})`;
const BARE_RANGE_RE = new RegExp(`^(?:${CELL}|${SIDE}:${SIDE})$`);
```

Please add `Jan`, `2024`, and `Q1` to the `splitRange` test at line 85 of `sheet-cells.test.ts` as sheet names, next to `Probe`. `Probe` passes today only because it has five letters.

src/tools/sheetCells.ts
```diff
@@ -0,0 +1,423 @@
...
+ *  grouped by sheet and the request order is preserved only WITHIN a sheet,
+ *  so each sheet gets its own cursor. */
+export function matchRangesToGridData(
+  ranges: string[],
+  sheets: SheetLike[],
+): { matches: RangeMatch[] } | { error: string } {
+  const byTitle = new Map<string, SheetLike>();
+  for (const sheet of sheets) {
+    const title = sheet.properties?.title;
+    if (typeof title === 'string') byTitle.set(title, sheet);
```
piotr-agier: Google resolves sheet titles case-insensitively, so `jan!A1` succeeds against a sheet named `Jan` and the API returns its data. This map is keyed on the exact title, so the lookup at line 185 misses and the tool reports `names sheet "jan", which is not in the response` for a request the API just fulfilled. Key the map and the cursors on a case-folded title, and keep the original title for `sheetTitle` in the output.

src/tools/sheetCells.ts
```diff
@@ -0,0 +1,423 @@
...
+        }
+        rowCells.push(out);
+        rowBytes += JSON.stringify(out).length + 1;
+      }
+
+      const wouldExceed =
+        usedCells + rowCells.length > budget.maxCells || usedBytes + rowBytes > budget.maxBytes;
+      // The first row always goes through: otherwise a range whose single row
+      // does not fit the budget would never advance and the client would loop.
+      if (wouldExceed && rowsConsumed > 0) { truncated = true; break; }
```
piotr-agier, line 309: The progress rule is per range, but the budget is per response, and that lets a multi-range read blow straight through it. Every range's first row is admitted regardless of what earlier ranges already consumed, and if that row is the whole range, `truncated` stays false and `exhausted` is never set.

Two cases I ran against this branch:
- Four ranges of one five-cell row each, `maxCells: 1`: returns all 20 cells, `truncated: false`, no `nextRanges`.
- A ten-cell range that exactly fills `maxCells: 10`, then a one-row range: returns 15 cells, `truncated: false`.

The guarantee should be "the response always makes progress", not "every range always makes progress". Admit an over-budget first row only when `usedCells === 0`. Otherwise, push the untouched range to `nextRanges` verbatim and set `exhausted`. The test at line 197 covers the single-range case only; please add the two above.

src/tools/sheetCells.ts
```diff
@@ -0,0 +1,423 @@
...
+        let present = false;
+        for (const field of fields) {
+          if (source[field] !== undefined) { out[field] = source[field]; present = true; }
+        }
+        if (!present) {
+          if (!budget.includeEmpty) continue;
+          out.empty = true;
+        }
+        rowCells.push(out);
+        rowBytes += JSON.stringify(out).length + 1;
```
piotr-agier: `.length` counts UTF-16 code units, not bytes. A cell of CJK text reported 102 here where the serialized output was 242 bytes, so a sheet of non-Latin text overruns `maxBytes` by more than 2x. Use `Buffer.byteLength(JSON.stringify(out)) + 1`, or rename the budget and the `returned.bytes` field to say what they measure.

src/tools/sheetCells.ts
```diff
@@ -0,0 +1,423 @@
...
+    if (requested.includes('dimensionGroups') && block.dimensionGroups === undefined) {
+      const shape = (groups: unknown[] | null | undefined) => (groups ?? []).map(raw => {
+        const group = raw as { range?: { startIndex?: number; endIndex?: number }; depth?: number; collapsed?: boolean };
+        return { start: group.range?.startIndex ?? 0, end: group.range?.endIndex ?? 0,
+                 depth: group.depth ?? 0, collapsed: group.collapsed === true };
+      });
+      block.dimensionGroups = { rows: shape(match.sheet.rowGroups), columns: shape(match.sheet.columnGroups) };
+    }
+
+    if (requested.includes('hiddenRows')) {
```
piotr-agier, line 431: Worth a comment here that `hiddenRows`, `hiddenColumns`, and `dimensionSizes` are collected from the grid's `rowMetadata` and `columnMetadata`, which the API scopes to the requested window, whereas `merges` above comes from the sheet-level list and covers the whole sheet. Both are reasonable, but the asymmetry should be stated in code and in the tool description (see the comment on the `sheetMetadata` property in `sheets.ts`), because an empty `hiddenRows` will otherwise be read as "this sheet has no hidden rows".

src/tools/sheetCells.ts
```diff
@@ -0,0 +1,423 @@
...
+  const { cellRange } = splitRange(match.range);
+  if (cellRange === null) {
+    // A bare whole-sheet reference: the original request had no end to reuse.
+    // Google returns rows spanning the sheet's real width, so the observed
+    // width is that width; the end row is left open (no digit), the same trick
+    // used for C:EB-style ranges below.
+    return `${prefix}${anchor}:${columnLetter(startColumn + Math.max(width, 1) - 1)}`;
+  }
+
+  const colon = cellRange.indexOf(':');
```
piotr-agier, line 247: For a row-only range such as `Alpha!6:9`, this produces `Alpha!A8:9`. I am not sure the Sheets API accepts a cell-to-row continuation like that. If it does not, paging a row range fails on the second call. Please check it live and either handle the row-only form here (rebuild both anchors as `A8:9` style rows, or as `A8:<width>9`) or add a test pinning whichever the API accepts.

src/tools/sheetCells.ts
```diff
@@ -0,0 +1,423 @@
...
+  sheetId: number;
+  sheetTitle: string;
+  grid: GridDataLike;
+  sheet: SheetLike;
+}
+
+/** Rebuilds the requested-range -> GridData correspondence. The response is
+ *  grouped by sheet and the request order is preserved only WITHIN a sheet,
+ *  so each sheet gets its own cursor. */
+export function matchRangesToGridData(
```
piotr-agier, line 176: One live case to verify: a requested range that is entirely empty. With this field mask, does the API return a `{}` element in `data` for it, or drop the element? If it drops it, the per-sheet cursor here goes out of step and every later range on that sheet is matched to the wrong grid, or fails with "fewer grids". The fixture has no empty range, so nothing covers it.

src/tools/sheets.ts
```diff
@@ -356,6 +372,35 @@ export const toolDefinitions: ToolDefinition[] = [
...
+        },
+        sheetMetadata: {
+          type: "array",
+          description: "Per-sheet extras returned once per sheet under `sheets`: merges (full extent, even when they leave the range), hiddenRows, hiddenColumns, frozen, dimensionGroups, dimensionSizes.",
+          items: { type: "string", enum: [...SHEET_METADATA] }
+        },
+        includeEmpty: { type: "boolean", description: "Return cells that have none of the requested fields as {a1, empty: true}. Default false - addresses are explicit, so gaps are unambiguous." },
+        maxCells: { type: "number", description: "Cell budget for the whole response (default 2000). On overflow the read stops at a row boundary and returns truncated:true plus nextRanges. On formula-heavy ranges the byte budget (maxBytes) normally binds first - a real spreadsheet of formula cells, reserialized into this tool's own compact per-cell JSON with the three default fields, measured at roughly 130 bytes per cell, exhausting the default maxBytes at around 1000 cells, still well before this default of 2000 - so truncation there is expected, not a sign of a misconfigured maxCells." },
```
piotr-agier: The `maxCells` and `maxBytes` descriptions repeat the same 130-bytes-per-cell paragraph, and every client pays for both in every `tools/list`. One sentence each is enough:

`maxCells`: "Cell budget for the whole response (default 2000). On overflow the read stops at a row boundary and returns truncated:true plus nextRanges."

`maxBytes`: "Byte budget for the serialized cells (default 131072), measured on this tool's output. With the default fields, formula-heavy ranges hit this before maxCells, at roughly 1000 cells."

src/tools/sheets.ts
```diff
@@ -356,6 +372,35 @@ export const toolDefinitions: ToolDefinition[] = [
...
+          type: "array",
+          description: "A1 ranges to read, e.g. [\"'Entity Assumptions'!B52:E52\", \"'Mgmt Fees Calc'!C6:EB13\"]. Returned separately, in this order, each with its own coordinates.",
+          items: { type: "string" }
+        },
+        fields: {
+          type: "array",
+          description: "CellData fields to return. Default ['userEnteredValue','effectiveValue','formattedValue'] gives the formula (userEnteredValue.formulaValue) together with its computed result (effectiveValue, or effectiveValue.errorValue for #REF!/#DIV/0!).",
+          items: { type: "string", enum: [...CELL_FIELDS] }
+        },
+        sheetMetadata: {
```
piotr-agier, line 392: Please say which of these are whole-sheet and which are window-scoped. Suggested: "Per-sheet extras returned once per sheet under `sheets`. `merges`, `frozen` and `dimensionGroups` cover the whole sheet (merges are reported at full extent even when they leave the range). `hiddenRows`, `hiddenColumns` and `dimensionSizes` cover only the rows and columns inside the requested ranges."

src/tools/sheets.ts
```diff
@@ -356,6 +372,35 @@ export const toolDefinitions: ToolDefinition[] = [
       required: ["spreadsheetId", "range"]
     }
   },
+  {
+    name: "getGoogleSheetCells",
```
piotr-agier, line 376: The description is long because it argues the case against `getGoogleSheetContent` three times. Once is enough, and a one-line rule helps the client more than the argument: "Use getGoogleSheetContent for a quick human-readable dump of one range. Use this tool whenever the result will be written back, compared to a formula, or spans several ranges."

test/tools/sheet-cells.test.ts
```diff
@@ -0,0 +1,398 @@
...
+  // Verify accumulation: both pixel sizes are present
+  assert.deepEqual(meta.Alpha.dimensionSizes, { columnWidths: { A: 100, G: 150 }, rowHeights: {} });
+});
+
+// This module's only test against a real, captured Sheets API response
+// (every other test above builds a small synthetic double instead) - runs
+// the actual pipeline (matchRangesToGridData -> extractRanges,
+// collectSheetMetadata) over test/fixtures/sheet-cells-grid.json and
+// asserts values read out of that fixture by hand (see the JSON directly,
+// or Task 1's report, for how each number below was derived).
```
piotr-agier: "Task 1's report" is not in the repo. Same for "the design doc" on line 384 and "verify the defect" on line 63. Please drop the references so the comments stand on their own. Line 63 can just say "Regression: the mask must have exactly one `sheets(...)` root."

test/tools/sheet-cells.test.ts
```diff
@@ -0,0 +1,398 @@
...
+test('the real captured spreadsheets.get fixture flows through matchRangesToGridData, extractRanges and collectSheetMetadata correctly', () => {
+  const fixturePath = path.join(process.cwd(), 'test', 'fixtures', 'sheet-cells-grid.json');
```
piotr-agier: This is the only test in the suite that resolves a file from `process.cwd()`, and `test/fixtures/` is a new directory. Resolving from `import.meta.url` keeps the test working when run from another directory:

```ts
const fixturePath = new URL('../fixtures/sheet-cells-grid.json', import.meta.url);
```

test/schema/tool-registry.test.ts
```diff
@@ -2,7 +2,7 @@ import assert from 'node:assert/strict';
 import { describe, it, before, after } from 'node:test';
 import { setupTestServer, type TestContext } from '../helpers/setup-server.js';
 
-const EXPECTED_TOOL_COUNT = 121;
+const EXPECTED_TOOL_COUNT = 122;
```
piotr-agier: This conflicts with master, which is at 128 after #209 and #211. After rebasing this should be 129, and the same in `README.md:9` and `docs/tools.md:3`. The `EXPECTED_TOOLS` entry at line 14 will need re-placing too, since that line has moved.

piotr-agier (changes requested):
Thanks, this closes a real gap: every Sheets write tool takes an absolute A1 range, and until now the only cell read reported range-relative row numbers with no column labels, so an agent could not safely write back what it just read. The helper split, the single-root field mask, and reporting merges at full extent are all the right calls, and the suite is green on the branch.

Two of the headline features have confirmed gaps that I have commented inline with repro cases: the bare-range grammar routes month and year tab names to the first sheet, and the budget is enforced per range rather than per response, so a multi-range read can return many times `maxCells` with `truncated: false`. Both are small fixes.

Two housekeeping items:
- Please rebase. Master moved to 128 tools and the three count lines conflict.
- Please add a CHANGELOG entry under `[Unreleased]` / `### Added` with the `**sheets:**` prefix and the PR link, matching the entries above it. That is the convention for every feature PR here.

Happy to re-review once those land.

src/tools/sheetCells.ts
salimaCodebase: Fixed in 01a1b05 with the construction you gave (COL/ROW/CELL/SIDE). `Jan`, `Dec`, `Log`, `Tax` and `2024` are now sheet names. Covered by a new `splitRange` test next to the `Probe` one, plus a companion test that the real A1 forms (`C6:E9`, `C:EB`, `6:9`, `A8:9`, `$A$1:$C$5`) still parse as ranges.

One note on `Q1`: the regex keeps a bare `Q1` a range, because it is a well-formed cell (column Q, row 1) and Google resolves it against the first sheet. Same for `FY25`. The test pins that, and pins `Q1!A1:C5` and `'Q1'` as the sheet name, which is how a Q1 tab has to be written in Sheets itself.

src/tools/sheetCells.ts
salimaCodebase: Fixed in 01a1b05. The lookup map and the per-sheet cursors are keyed on `title.toLowerCase()`; `sheetTitle` in the output now comes from `sheet.properties.title`, so it carries the sheet's own spelling. Error messages keep the requested spelling. Test: `jan!A1` + `JAN!A10` against a sheet named `Jan` resolve to `Jan` and take the first and second grid, so the shared cursor is pinned too.

src/tools/sheetCells.ts
salimaCodebase: Reproduced both cases against the branch before the fix: four ranges of one five-cell row each at `maxCells: 1` returned 20 cells, `truncated: false`, `nextRanges: []`; the ten-cell range at `maxCells: 10` followed by a one-row range returned 15 cells, `truncated: false`.

Fixed in 01a1b05: an over-budget first row is admitted only while `usedCells === 0`. Otherwise the range is left untouched, pushed to `nextRanges` verbatim, and `exhausted` is set, so it gets no result block either, exactly like a range the loop never reached. Both cases are now tests: the first returns 5 cells with `nextRanges` holding the other three ranges verbatim, the second returns 10 cells and one result block.

src/tools/sheetCells.ts
salimaCodebase: Fixed in 01a1b05: `Buffer.byteLength(JSON.stringify(out)) + 1`. Two tests: one asserts `returned.bytes` equals the summed `Buffer.byteLength` of the emitted cells and exceeds the code-unit count for the same content, the other sets `maxBytes` to a value that holds two rows of CJK measured in code units but only one measured in bytes, and pins that the read truncates after the first row with `nextRanges: ['Alpha!A2:A3']`.

src/tools/sheetCells.ts
salimaCodebase: Added in b774179, above the `hiddenRows` block: `hiddenRows`/`hiddenColumns`/`dimensionSizes` come from the grid's `rowMetadata`/`columnMetadata`, which the API scopes to the requested window, while `merges`/`frozen`/`dimensionGroups` come from the sheet-level lists and cover the whole sheet — so an empty `hiddenRows` means "none inside the ranges that were read". The `sheetMetadata` description now says the same thing.

src/tools/sheetCells.ts
salimaCodebase: Checked live on a probe spreadsheet: the API does accept the cell-to-row form. `Probe!A8:9` came back with `startRow: 7`, `startColumn` omitted, and 2 rows of `rowData` — so the anchor `remainderRange` builds for a row-only request is valid and paging works on the second call. No code change; pinned by a test in b774179 that `Alpha!6:9` continues as `Alpha!A8:9` and that the continuation round-trips back through the bare-range grammar as `{ sheetTitle: 'Alpha', cellRange: 'A8:9' }`.

src/tools/sheetCells.ts
salimaCodebase: Checked live with this field mask: the element is returned, not dropped. Requesting `Probe!Z90:AA95` (entirely empty) and `Probe!A1:B2` in one call returned two `data[]` elements — the first with `startRow: 89`/`startColumn: 25` and no `rowData`, the second with `rowData` and both offsets elided because they are 0. The per-sheet cursor therefore cannot go out of step. Test added in b774179 over exactly that shape: the empty range yields an empty `cells` array in its own slot, and the second range addresses `A1`, not `Z90`.

src/tools/sheets.ts
salimaCodebase: Replaced in b774179 with your two sentences, verbatim.

src/tools/sheets.ts
salimaCodebase: Replaced in b774179 with your wording, verbatim. `docs/tools.md` carries the same split.

src/tools/sheets.ts
salimaCodebase: Cut in b774179: the three-way argument is gone, replaced by your one-line rule.

test/tools/sheet-cells.test.ts
salimaCodebase: All three dropped in b774179. Line 63 is now your text. The fixture preamble says the values were hand-read from the checked-in JSON and can be confirmed against it directly; the line 384 comment now explains the nonzero-`startRow` case itself (a range-relative reading would report those cells six rows too high).

test/tools/sheet-cells.test.ts
salimaCodebase: Changed in b774179 to the form you gave. One extra change was needed: `test:build` compiles into `.tmp-test/`, which only received `.ts` output, so `import.meta.url` resolved to a directory with no fixture. `test:build` now copies `test/fixtures` alongside the compiled tests.

test/schema/tool-registry.test.ts
salimaCodebase: Rebased onto master (now at `0490ea2`, after #209 and #211). Confirmed 129 by running the registry test rather than by arithmetic: `EXPECTED_TOOL_COUNT`, `README.md:9` and `docs/tools.md:3` all read 129. `EXPECTED_TOOLS` merged cleanly — `getGoogleSheetCells` sits on the sheets line next to `getGoogleSheetContent`.

Full gate green on the rebased branch: `npm ci && npm run lint && npm run build && npm run registry:validate && npm test` — 1069 tests, 0 failures.

piotr-agier:
One thing left before this can go in: **please rebase onto current master (`d093760`).**

The green checks on this PR ran on 2026-09-17 against base `0490ea2`, and master has moved since. GitHub still reports the PR mergeable because the conflict is semantic rather than textual. I test-merged current master locally and the suite fails:

```
not ok 222 - Documentation reference
  → never names a tool that is not registered
# tests 1080  # pass 1079  # fail 1
```

#215 (`test: fail the build when a tool description names a tool that does not exist`) landed after your base, and it flags 14 identifiers in the `getGoogleSheetCells` description as tools that do not exist:

`dataValidation`, `dimensionGroups`, `dimensionSizes`, `effectiveFormat`, `effectiveValue`, `errorValue`, `formattedValue`, `formulaValue`, `hiddenColumns`, `hiddenRows`, `nextRanges`, `textFormatRuns`, `userEnteredFormat`, `userEnteredValue`

These are all real Sheets API field names plus one output field of this tool, so this is exactly the "extend deliberately" case the gate's comment describes. The gate exempts a tool's own *parameters* by walking `properties`, which does not reach these: 11 of them are `enum` members of `fields`/`sheetMetadata`, and `nextRanges`/`errorValue`/`formulaValue` are output names. (`note`, `hyperlink`, `merges` and `frozen` escape only because they are not camelCase.)

Please add them to `NOT_TOOL_NAMES` in `test/docs-reference.test.ts`, in the existing groups:

```ts
      // Google API field names
      'byteLength', 'contentUri', 'createdTime', 'dataBase64', 'externalOnly',
      'fileOrganizer', 'fullText', 'hangoutsMeet', 'hiddenByUser', 'horizontalRule',
      'mimeType', 'modifiedTime', 'nextOffset', 'objectId', 'revisionId', 'startTime',
      // Sheets CellData, ExtendedValue and dimension-metadata field names
      'dataValidation', 'dimensionGroups', 'dimensionSizes', 'effectiveFormat',
      'effectiveValue', 'errorValue', 'formattedValue', 'formulaValue',
      'hiddenColumns', 'hiddenRows', 'textFormatRuns', 'userEnteredFormat',
      'userEnteredValue',
      // Output field names, named in prose by the tool that returns them
      'nextRanges',
```

salimaCodebase:
Rebased onto current master, `c61b1a6` ("chore: prepare release 2.11.0 (#221)"), not the `d093760` you named — master moved while this was open and `d093760` is now an ancestor of `c61b1a6`, so rebasing onto it would only have needed doing again. New head: `04741c3`.

The 14 names are in `NOT_TOOL_NAMES` in the grouping you gave, with your two comment headings verbatim: the 13 Sheets field names under "Sheets CellData, ExtendedValue and dimension-metadata field names", and `nextRanges` under "Output field names, named in prose by the tool that returns them".

One conflict to note: the changelog entry. `2.11.0` shipped the `## [Unreleased]` block this branch was written against, so the `getGoogleSheetCells` bullet moved into a fresh `### Added` under `## [Unreleased]`; the released section is untouched. The tool counters did not conflict — master is still at 128, so `EXPECTED_TOOL_COUNT`, `README.md:9` and `docs/tools.md:3` all read 129, confirmed by running the registry test rather than by arithmetic.

Full gate green on the rebased branch: `npm ci && npm run lint && npm run build && npm run registry:validate && npm test` — 1081 tests, 0 failures, `Documentation reference` included.

PR #217, opened 2026-09-17
feat(sheets): updateGoogleSheetIfUnchanged — guarded writes with a rollback pre-image

salimaCodebase:
> **Stacked on #210 — please read first.** This branch is cut from the head of
> #210 (`feat/sheets-structured-cell-read`), not from `master`, because the new
> code imports helpers that exist only there: `cellA1`, `splitRange`,
> `buildFieldMask`, `matchRangesToGridData`, `extractRanges` and the `RangeMatch`
> / `SheetLike` types from `src/tools/sheetCells.ts`, a file `master` does not
> have yet. A branch cut from `master` would not build, and duplicating
> `sheetCells.ts` into this PR would just make two PRs adding the same file
> conflict with each other. **So the diff below shows both changesets — #210's
> `getGoogleSheetCells` and this tool — until #210 merges, at which point it
> shrinks by itself to the guarded-write commits alone.** CI here runs the merge
> of this head into `master`, so it carries #210's commits and builds. Review
> only the two commits titled `feat(sheets): canonical projection…` and
> `feat(sheets): updateGoogleSheetIfUnchanged…` (plus the CHANGELOG commit);
> everything else is #210, already under review there.

## What did not work

Every Sheets write in this server is unconditional. `updateGoogleSheet` and
`batchUpdateGoogleSheetValues` send values and overwrite whatever is there, and
the reply reports counts, not contents. If the numbers moved between the read
that produced them and the write that acts on them, the write wins silently —
and what was overwritten is gone, because nothing in the response can restore
it.

The Docs side of this server does not have that problem: `ifRevisionId` rides on
the Docs `batchUpdate` as `WriteControl.requiredRevisionId` and the API itself
refuses a stale write (#187). Sheets has no equivalent. That is not an
assumption carried over from the Docs work — it was established by experiment
against the live API before any of this was designed:

- `spreadsheets.batchUpdate` carrying a deliberately **wrong**
  `requiredRevisionId` **applies anyway**. The field is accepted and ignored.
- `spreadsheets.get` exposes **no revision field at all**, so there is nothing
  to send even if it were honoured.

So there is no compare-and-swap to lean on, and the check has to live in the
connector.

## Why it matters

A spreadsheet is the document type most likely to be open in a colleague's
browser while an agent edits it, and its cells are the content where a silent
overwrite is hardest to notice afterwards — a changed number looks exactly like
a number. "Read the range, decide, write it back" is the normal shape of the
work, and the gap between the read and the decision is where a human edit lands.

## What changed

Adds **`updateGoogleSheetIfUnchanged`** (tool count 129 → 130), a write tool
that refuses to write when the guarded area is not what the caller last saw, and
reports what it overwrote when it does write.

- Call once with `dryRun: true` to get a fingerprint of the guarded area, then
  again passing it back as `expectedFingerprint`. A mismatch is refused **with
  the area's current contents attached**, so the caller can see what moved
  instead of guessing.
- `guardRanges` defaults to the ranges being written, and need not cover them —
  guarding a source block while writing a summary elsewhere is a valid pattern,
  though then the guard says nothing about what is being overwritten, and the
  description says exactly that.
- The response carries `preImage`, a `{range, values}` payload in precisely the
  shape `updates` accepts, plus `postFingerprint`. **The undo path is this same
  tool applied to its own returned pre-image**: pass `preImage` back as
  `updates` and `postFingerprint` as `expectedFingerprint`, and the undo is
  itself guarded. No rollback state is kept anywhere — not in this server, not
  on disk, not in the spreadsheet. Everything an undo needs is in the response
  the caller already holds.
- `hazards` lists cells that cannot be restored through `updates`, which carries
  strings and would silently change their type on the way back (text kept as
  text behind a leading apostrophe that reads as a formula, a number, or
  TRUE/FALSE). Each hazard comes with the cell's native `userEnteredValue`, to
  be applied through `sheetsBatchUpdate`'s `updateCells` instead. The list is
  deliberately **not** exhaustive and says so: date-shaped text is not detected,
  because Google resolves a date from the locale and the cell format, not from
  the string alone.

**Optimistic, not atomic — and the tool description says so in those words**
rather than implying a lock. Because the API has no compare-and-swap, the
comparison happens in this server, so a write landing in the gap between the
read and the write — well under a second — is not caught. What is caught is the
case that actually happens: the sheet changed *before* that gap, between the
caller reading it and deciding what to write. A guarantee described honestly is
worth more than one implied and then quietly broken.

Two API calls, not three. The guard read doubles as the pre-image read, and the
post-write fingerprint is computed from the write response
(`includeValuesInResponse` with `FORMULA` rendering) rather than from a second
read; when the guard set equals the write set — the default — one read serves
both. The post-write block is built from `updatedData.range`, which echoes the
declared written range in full (checked live: writing one value into a declared
3×3 range returns that same 3×3, with `values` trimmed to 1×1), so it rests on
the same geometry the guard fingerprint did. If the write succeeds but its
response cannot be fingerprinted, the tool says so and hands back the pre-image
rather than inventing a `postFingerprint` a later rollback would trust.

The comparison logic lives in `src/tools/sheetGuard.ts`, apart from the API
call, so it is unit-testable without stubbing googleapis — the same split
`sheetCells.ts` and `batchPassthrough.ts` already use. Both sides are projected
into one canonical space before anything is compared (`spreadsheets.get` returns
`CellData`, a write response returns a rendering), and every block is padded out
to its **declared** A1 rectangle, since Google trims empty edges on both the
read and the write-response side — without that, a cleared cell vanishes from
the pre-image instead of coming back as an empty string that clears it again.
The fingerprint covers each block's sheet title, origin and dimensions as well
as its cells, so inserting a row inside the guarded area changes it even when
every surviving cell still holds what it held; the title goes through
`JSON.stringify` because Google accepts a sheet title containing a literal
newline, which raw interpolation would let forge a collision.

Verified against a real spreadsheet as well as in unit tests: a guarded write
applying, the same write refused after an outside edit, and a full round trip
undone through the returned pre-image.

## Backward compatibility

Purely additive. No existing tool, schema, output shape or shared helper
changes; the diff adds lines to `src/tools/sheets.ts` and removes none.
`updateGoogleSheet` and `batchUpdateGoogleSheetValues` are untouched and remain
the right tools when no guard is wanted. The new tool needs no scope beyond the
Sheets write scopes `updateGoogleSheet` already uses. README, `docs/tools.md`
and `test/schema/tool-registry.test.ts` move 129 → 130 together.

`npm ci && npm run lint && npm run build && npm run registry:validate && npm
test` is green on this branch: 1112 tests, 0 failures.


salimaCodebase:
Updated the branch (28ceec7) with two correctness fixes found by a review after this PR was opened:

1. `declaredRectangle` borrowed `convertA1ToGridRange`, which fabricates a missing end bound instead of reporting one is missing. `Data!A2:C` and `A5:A` were treated as 1-row rectangles, so a guard on a multi-row range fingerprinted only its first row, and the pre-image for a multi-row write captured only that first row too — an automated rollback would have restored one row and left the rest of the overwrite in place.
2. The same helper's regex is uppercase-only and rejects `$`, so `a1:c3` and `A$1:C$3` fell back to the trimmed observed extent while the post-write block (built from Google's canonical uppercase echo) did not — permanently mismatching and refusing the documented rollback for those spellings.

Both are fixed with a strict, self-contained rectangle parser in `sheetGuard.ts` that normalizes case/`$` and only reports a rectangle when both anchors are fully bounded, falling back to the observed extent otherwise (unchanged behavior for genuinely open-ended ranges).

The rest of the push is performance work on the same code path found in the same review: a cap on the declared rectangle so an unbounded-looking range like `A1:Z100000` can't materialize millions of cells, streaming the fingerprint hash instead of building and joining an array, avoiding a duplicate block computation on the default guard-equals-write path, computing `guardContents` lazily, skipping an unused read on `dryRun` with distinct guard ranges, and preserving `preImage`/`hazards` when the write call itself throws after being sent.

Test suite: 1124 passing (up from 1112).

piotr-agier (commented):
Reviewed only the four guarded-write commits on top of #210's head (acafc16, 7b1fb38, 28acfcc, 28ceec7), as the description asks. The design is sound and honestly documented; lint, build and the full suite (1124 tests) pass on this head locally, and CI is green on Node 22 and 24.

One blocking item: write ranges without a bounded declared rectangle (open-ended, whole-sheet, reversed anchors) silently break the documented undo path. Details inline, plus a few nits.

Merge order: #210 first. This diff then shrinks to the four commits above on its own.

src/tools/sheets.ts
```diff
@@ -1027,6 +1125,201 @@ export async function handleTool(
...
+      // declared guard/write rectangle is absurdly large - see
+      // MAX_DECLARED_RECTANGLE_CELLS in sheetGuard.ts - rather than
+      // materializing it. That is a plain, user-facing "narrow your range"
+      // condition, exactly like the matchRangesToGridData `{error}` case
+      // just below, so it is handled locally the same way rather than left
+      // to bubble up as a bare thrown exception.
+      try {
+
+      const guardRanges = guardRangesFor(a.updates, a.guardRanges);
+      const writeRanges = a.updates.map(u => u.range);
```
piotr-agier, line 1145: **Blocking: write ranges without a bounded rectangle break the undo contract**

Every range in `updates` needs a bounded declared rectangle for the rest of this handler to keep its promises, and nothing checks that here. `declaredRectangle` returns null for `Probe!A2:C`, `A:C`, `5:9`, a bare `Probe`, and reversed anchors, and the block then falls back to the observed extent. That is fine for a guard-only range, since dryRun and the write read it the same way. For a write range it fails in two ways:

1. The pre-image under-covers. A read of `Probe!A2:C` is trimmed to existing data. If the sheet holds 2 rows there and the update writes 5, `preImage` has 2 rows, so applying it restores 2 rows and leaves 3 rows of overwritten output in place. Same class as the `A1:C` bug fixed in 28ceec7, just on the write side.
2. `postFingerprint` never matches a re-guard. The post block is built from `updatedData.range`, which is the bounded rectangle Google actually wrote. The undo call's guard read of `Probe!A2:C` observes down to the last data row on the sheet, which is a different geometry, so the undo is refused with `fingerprint-mismatch` even though the description promises it works.

Suggest validating up front that each write range parses to a bounded rectangle and refusing otherwise with a "narrow the range" message, in the same style as the cell-cap error. Guard ranges can stay as they are. Please add a handler test with an open-ended write range that pins the refusal.

src/tools/sheets.ts
```diff
@@ -1027,6 +1125,201 @@ export async function handleTool(
...
+          }) }],
+          isError: true
+        };
+      }
+
+      const postBlocks: CanonicalBlock[] = writeResponses.map((r, i) => {
+        const updatedData = r.updatedData!;
+        const values = (updatedData.values ?? []) as unknown[][];
+        const cells = values.map(row => row.map(v => projectResponseValue(v)));
+        return padToDeclaredRange(updatedData.range!, writeMatches[i]?.sheetTitle ?? '', cells,
```
piotr-agier, line 1303: **Follow-on to the write-range comment: a throw after the write drops `preImage`**

This call can throw after the write has already landed: with an open-ended write range the echoed rectangle can exceed `MAX_DECLARED_RECTANGLE_CELLS`, and the outer catch at line 1318 turns that into a bare `errorResponse`, discarding `preImage` and `hazards` on exactly the path where they matter. Rejecting unbounded write ranges up front removes the case, but it is worth either narrowing the outer try so post-write throws cannot reach it, or wrapping this block the way the `batchUpdate` call is wrapped at line 1239 so a failure here still hands back `preImage`.

src/tools/sheetGuard.ts
```diff
@@ -0,0 +1,329 @@
...
+
+  const [, startColLetters, startRowDigits, endColLetters, endRowDigits] = match;
+  const startColumn = colToIndex(startColLetters);
+  const startRow = parseInt(startRowDigits, 10) - 1;
+  const endColumn = endColLetters ? colToIndex(endColLetters) + 1 : startColumn + 1;
+  const endRow = endRowDigits ? parseInt(endRowDigits, 10) : startRow + 1;
+
+  const rows = endRow - startRow;
+  const columns = endColumn - startColumn;
+  if (rows <= 0 || columns <= 0) return null; // reversed anchors: no rectangle to pad to
```
piotr-agier, line 247: **Nit: reversed anchors could be normalized instead of dropped**

Google normalizes `C3:A1` to `A1:C3`, so the write echo comes back bounded while the guard side falls back to the observed extent, and the two geometries diverge. Taking `min`/`max` of the two anchors here would make `C3:A1` and `A1:C3` the same rectangle on both sides, which is what the `$` and lowercase normalization above already does for spelling. Cheap, and it removes one more input from the "no rectangle" set.

src/tools/sheets.ts
```diff
@@ -1027,6 +1125,201 @@ export async function handleTool(
...
+          spreadsheetId: a.spreadsheetId,
+          written: { ranges: writeRanges.length, cells: written.data.totalUpdatedCells ?? 0 },
+          fingerprintVerified: fingerprint,
+          postFingerprint: fingerprintOf(postBlocks),
+          preImage,
+          hazards
+        }) }],
+        isError: false
+      };
+      } catch (err) {
```
piotr-agier, line 1318: **Nit: the local catch-all hides the dispatcher's log line**

This catches everything thrown between lines 1142 and 1318, including Google API errors from `spreadsheets.get`, and returns the same `errorResponse(message)` the dispatcher in `index.ts` would, except the dispatcher also calls `log(...)`. The comment at line 1133 says the try exists for the cell-cap error from `padToDeclaredRange`. Either scope the try to the two `blockFromMatch` sites (lines 1194 and 1230), or rethrow anything that is not the cap error so API failures keep their log entry.

src/tools/sheets.ts
```diff
@@ -338,6 +374,39 @@ export const toolDefinitions: ToolDefinition[] = [
...
...it, and get back what was overwritten so the change can be undone. Call once with dryRun:true to obtain the fingerprint, then again passing it as expectedFingerprint. IMPORTANT: this is optimistic and NOT atomic. The Sheets API has no compare-and-swap - unlike Docs, where ifRevisionId is enforced by the API itself - so the check happens in this server, and a write landing in the gap between the read and the write (well under a second) is not caught. It catches the case that actually happens: the model changed since you last looked. To undo a write, call again with updates set to the returned preImage, guardRanges set to the ranges you wrote, and expectedFingerprint set to the returned postFingerprint. Any cell listed in the returned hazards cannot be restored through updates, which only accepts strings and would silently change the cell's type on rollback - apply its returned userEnteredValue verbatim through another tool (e.g. sheetsBatchUpdate's updateCells) instead. The hazard list is deliberately not exhaustive: text kept as text via a leading apostrophe is only flagged when it reads as a formula, a number, or TRUE/FALSE - date-shaped text (e.g. '2024-01-01') is not detected.",
```
piotr-agier: **Nit: undo instructions ask for a redundant parameter**

In the sentence beginning "To undo a write, call again with updates set to the returned preImage, guardRanges set to the ranges you wrote": `guardRanges` already defaults to the ranges in `updates`, and `preImage` carries those exact ranges, so an undo needs only `updates: preImage` and `expectedFingerprint: postFingerprint`. Asking callers to also pass `guardRanges` adds a step they can get wrong (a typo there produces a mismatch refusal that looks like a real conflict). Suggest dropping it from the description, or saying it is optional.

test/tools/sheet-guard-handler.test.ts
```diff
@@ -0,0 +1,475 @@
...
+  rowDataReads = 0;
+
+  const { ctx } = fakeCtx(grid, { writeResponse: writeOk });
+  const result = await handleTool('updateGoogleSheetIfUnchanged',
+    { spreadsheetId: 'x', updates: [{ range: 'Probe!A1', values: [['1']] }], expectedFingerprint: fp }, ctx);
+  assert.equal(result!.isError, false);
+  assert.equal(rowDataReads, 1, 'rowData must be read exactly once - not twice for the same guard/write match');
+});
+
+test('an absurdly large declared guard range is refused with a clear error, not materialized (finding C)', async () => {
```
piotr-agier, line 226: **Nit: test comments reference an external review's labels**

"finding C", "finding D", "finding E", "finding F", "finding G" and "finding 1/2" (lines 126, 202, 226, 242, 266) refer to a review that is not in this repo, so they will mean nothing to the next reader. The test names already say what is pinned; suggest deleting the labels and keeping the behavioural description.

salimaCodebase:
Rebased onto the new head of #210 (`04741c3`), since this branch is stacked on it and #210 was rebased onto current master `c61b1a6`. New head here: `37b83fd`.

The #215 tool-name gate flagged four identifiers in the `updateGoogleSheetIfUnchanged` description. `preImage` and `postFingerprint` are output fields this tool returns and names in prose, so they joined the "Output field names" group #210 adds. The other two were a real phantom rather than a false positive: the hazard advice pointed at `sheetsBatchUpdate`'s `updateCells`, and no tool by that name is registered here, so a model following it would spend a lookup finding nothing. `docs/tools.md` already said "through another tool" without naming one; the description now matches it. `hazards` escapes the gate only because it is not camelCase.

Counters: 130, from the registry test. Full gate green — `npm ci && npm run lint && npm run build && npm run registry:validate && npm test`, 1136 tests, 0 failures.

piotr-agier:
This needs another rebase + conflict resolution, please, @salimaCodebase

src/tools/sheets.ts
salimaCodebase: Fixed. `checkWriteRanges` in sheetGuard.ts now runs over every range in `updates` before the read, and refuses any that does not parse to a bounded rectangle with a "narrow the range" message in the same style as the cell-cap error; it also surfaces the MAX_DECLARED_RECTANGLE_CELLS cap for write ranges at the same point, so a request that cannot be honoured costs no API call. Guard ranges are unchanged and may stay open-ended. Handler tests pin the refusal for `Probe!A2:C`, `Probe!A5:A`, `Probe!A:C`, `Probe!5:9` and a bare `Probe`, on the write path and on dryRun, asserting zero reads and zero writes, plus one that an open-ended guard range is still accepted.

Also: #210 merged, so the branch is no longer stacked on it - it is rebased onto master directly (now 29b1172).

src/tools/sheets.ts
salimaCodebase: Took both halves. The post-write block is now wrapped the way the `batchUpdate` call is: a throw there returns `error: "post-write-fingerprint-unavailable"` with `written`, `preImage` and `hazards`, the same shape the incomplete-response branch already used. The outer try is gone as well (see the reply on the catch-all comment), so nothing post-write can reach it either way. Test forces it with an echoed `updatedData.range` past the cell cap and asserts `preImage` comes back.

src/tools/sheets.ts
salimaCodebase: Fixed. The try is now scoped to the two `blockFromMatch` passes alone, via a small `blocksFor` helper that returns `{blocks}` or `{error}`; everything else, `spreadsheets.get` included, propagates to the dispatcher and keeps its `log(...)`. Test asserts `handleTool` rejects on a thrown API error rather than returning an `errorResponse`.

src/tools/sheets.ts
salimaCodebase: Fixed. The sentence now reads "call again with updates set to the returned preImage and expectedFingerprint set to the returned postFingerprint - nothing else, since guardRanges defaults to the ranges in updates and preImage already carries exactly the ranges that were written." The round-trip test now performs its undo without `guardRanges` too.

src/tools/sheetGuard.ts
salimaCodebase: Done. The two anchors are treated as corners and normalized by min/max, so `C3:A1` is the same rectangle as `A1:C3`; the `rows <= 0 || columns <= 0` branch is gone with it, since neither can happen now. Tests cover `C3:A1`, the row-reversed `A3:C1` and the column-reversed `C1:A3`, each observing one cell so the padding is what is pinned rather than two equal-sized blocks agreeing by accident, plus `C5:A3` for the origin.

test/tools/sheet-guard-handler.test.ts
salimaCodebase: Done, and swept the rest of the diff. Removed the labels from both test files (the handler test and sheet-guard.test.ts, which carried "Finding A/B/C" section headers) and from a "see finding A below" in the sheetGuard.ts comment block, keeping the behavioural description in each case. One more of ours had leaked into the CHANGELOG: it named a tool that does not exist in this repo, now reworded to the underlying `spreadsheets.batchUpdate` `updateCells` request.

src/tools/sheets.ts
salimaCodebase: Unprompted, on the same description: the hazard sentence told callers to apply the returned `userEnteredValue` "through another tool". Nothing this server registers can do that - there is no raw batchUpdate passthrough and no tool that writes a native `userEnteredValue` - so the advice named an action a caller cannot perform. Pre-existing on this branch, found while working through your comments rather than raised by one of them.

It now says the cell cannot be restored by feeding `preImage` back, since `updates` carries strings and re-writing the string re-interprets it; that the native `userEnteredValue` is returned so the one cell can be restored outside this server, through the Sheets API's own `updateCells` or by hand; and that an undo touching a hazard cell needs that cell handled separately. No tool is named. Changed in the description and in the `docs/tools.md` entry, which duplicated the prose, with a documentation-reference test pinning the shared sentence in both so they cannot drift apart again.

Issue #169, opened 2026-08-16
Token refresh is unbounded — a stalled refresh hangs every subsequent call on the server


shopology:
`AccountClientFactory.refreshIfNeeded()` awaits `client.refreshAccessToken()` with no timeout, abort or retry. If the token endpoint stalls, the await never settles — and because `inflightRefresh` dedupes, every subsequent call awaits the same pending promise, so the whole server appears dead until it eventually resolves.

`src/auth/accountClientFactory.ts`:

```js
const { credentials } = await client.refreshAccessToken();
```

Two things make it hard to diagnose. `REFRESH_BUFFER_MS` fires the refresh five minutes *before* expiry, so the stall lands on an ordinary unrelated call with no apparent connection to auth. And a restart appears to fix it, because it resets the token clock and moves you away from the boundary.

**Observed:** twice in one session over stdio — 4+ minute hangs, no error raised, self-recovering, while an independent read path to the same Google account kept working throughout. Token mint times confirmed the refresh was due within a minute of each hang starting. The hangs also exceeded the 120s `--api-timeout`, which independently suggests this path bypasses the timeout wrapper.

**Suggested fix:** route it through the existing `withRetry` in `src/utils/retry.core.js`, which already has the AbortController, configurable timeout and backoff, and retries on ETIMEDOUT/ECONNRESET/429/503/504. A tighter timeout than the 120s global default seems appropriate for a token endpoint. `invalid_grant` isn't in the retryable set, so a revoked grant would still surface immediately with the existing actionable message.

I've been running a patched build locally for this and tested it against a simulated stalled endpoint: unpatched hangs indefinitely, patched raises TimeoutError after the configured attempts, a transient stall recovers silently on retry, and `invalid_grant` surfaces on the first attempt without retrying. Live, it's since crossed a real refresh boundary cleanly.

Happy to open a PR if that's useful — let me know and I'll tidy it up against your conventions rather than assume them.

One note: `src/auth/team/clientFactory.ts` has the same unbounded call and would presumably want the same treatment.

Thanks for the project — it's become a core part of my daily workflow.

PR #218, opened 2026-09-18
feat: apply the existing withRetry timeout/retry policy to every Google API call site

salimaCodebase:
## What

Applies the retry/timeout helper this repo already ships — `withRetry` in `src/utils/retry.ts` — to every Google API call site in the five tool files. No new mechanism, no new configuration, no behaviour change on a healthy call: 176 call sites that previously awaited Google with no deadline now run under one.

The rule is one line, and it splits reads from writes:

- **Reads** get the full policy: `withRetry(fn, ctx.runtimeConfig, label, ctx.log)` — a per-attempt deadline plus the existing bounded backoff. Repeating a read is safe.
- **Writes, and anything else with a side effect,** get the deadline **only**: `{ ...ctx.runtimeConfig, retryMax: 0 }`. They gain hang protection without gaining a retry they never had.

| file | reads (full retry) | writes (timeout only) |
|---|---:|---:|
| `src/tools/calendar.ts` | 4 | 3 |
| `src/tools/slides.ts` | 12 | 24 |
| `src/tools/sheets.ts` | 12 | 19 |
| `src/tools/drive.ts` | 27 | 23 |
| `src/tools/docs.ts` | 31 | 21 |
| **total** | **86** | **90** |

## Why

`withRetry` was written for exactly this and was almost entirely unused: across ~176 API call sites in the five tool files it appeared **once**, in `docs.ts`. Everywhere else — every `documents.batchUpdate`, every `files.get`, every `spreadsheets.values.update` — the handler awaited a bare googleapis promise with no timeout and no cancellation.

A stalled Google response therefore held the MCP request open indefinitely. On a stdio deployment that is a hung tool call the user eventually notices; on a hosted one it is worse, because the stalled request also holds a server concurrency slot shared with every other consumer. This is the same failure the token-refresh work in 2.8.0 fixed for OAuth refresh (#169) — the tool call sites were simply never given the same treatment.

## Writes deliberately do **not** gain retries

This is the question a 176-site diff should raise first, so to be explicit: **no call that was not retried before is retried now.**

The policy in `src/utils/retry.core.js` treats `ECONNRESET` (and undici's `UND_ERR_SOCKET`) as retryable transport errors. A reset can be raised *after* a server has accepted and begun processing a request, so a blind retry of a non-idempotent write can double-apply it — a second `insertText`, a second `events.insert`, a duplicated row. The same file already excludes HTTP 500 and 502 from the retryable set for exactly this reason (its comment names `documents.batchUpdate` by name); `retryMax: 0` extends that caution from the status codes to the transport codes for the write path.

So writes get the deadline — which does real work, because `withRetry` aborts the attempt through the `AbortSignal` it hands `fn`, and every wrapped call forwards it as gaxios `{ signal }`, so a timed-out request is actually cancelled rather than left in flight — and nothing else.

Two classification notes:

- `spreadsheets.get`, `documents.get` and friends are reads even when a write tool calls them to resolve a sheet id, a grid range or a `textToFind` index. They are wrapped as reads.
- The one pre-existing `withRetry` call, `createGoogleDoc`'s initial-content `batchUpdate`, is a write with full retries. It is left exactly as it was — that was an upstream choice that predates this PR, and changing it is not this PR's business.

## What is not wrapped, and why

Ten call sites are deliberately left alone, all of them in shared helper modules outside the five tool files rather than call sites those files own:

- `src/utils/driveImageUpload.ts` — `uploadImageToDrive` (3) and `deleteDriveFile` (1)
- `src/tools/text-content.ts` — `downloadTextContent`, `writeTextContent` (2)
- `src/download-file.ts` — `downloadDriveFile` (3)
- `src/auth/identity.ts` — `getEffectiveIdentity` (1)

`manage_accounts` delegates entirely to `ctx.accountOps` and makes no direct Google API call in `drive.ts`. Extending the same rule to those modules is a natural follow-up; keeping this PR to the five tool files keeps the diff mechanical and reviewable.

The two `{ responseType: 'stream' }` reads in `restoreRevision` *are* wrapped, with full retries. A stream-mode request's promise settles when response headers arrive, not when the body is read — both call sites only build a Node `Readable` after the await resolves — so the deadline bounds the time to obtain the stream, not an in-progress transfer.

## Backward compatibility

No tool, parameter, or configuration option is added, removed or renamed. The knobs already exist and already have defaults: `--api-timeout` / `GOOGLE_DRIVE_MCP_API_TIMEOUT` (120s), `--retry-max`, `--retry-base-delay`. A healthy call behaves exactly as before. The only observable change is that a call which would previously have hung forever now fails with a clear `… timed out after Nms`, and a read that hits a 429/503/504 or a transport blip now succeeds where it previously surfaced the error.

Setting `--api-timeout=0` restores the old unbounded behaviour everywhere, if anyone wants it.

## How tested

Four integration tests in `test/integration/api-timeouts.test.ts`, which is what makes the read/write split an enforceable rule rather than a convention — `test/retry.test.ts` covers `withRetry` itself, this covers the policy as applied at the call sites:

1. a read (`calendar.events.list`) is retried past a 503 and succeeds;
2. a write (`calendar.events.insert`) fails on the first 503 without retrying;
3. the same write still fails with a timeout when the call never settles;
4. a stream-mode read (`drive.revisions.get`, standing in for `restoreRevision`'s binary-file fetch) is retried like any other read.

The file sets a 100ms `apiTimeout` and a 5ms `retryBaseDelay` before the first dynamic import of the server module; Node's test runner gives each file its own process, so this does not leak into the other suites.

Full gate green on this branch: `npm ci && npm run lint && npm run build && npm run registry:validate && npm test` — 1027 tests, 150 suites, 0 failures. Run locally on Node 22.


src/tools/docs.ts
```diff
@@ -4105,10 +4240,15 @@ export async function handleTool(toolName: string, args: Record<string, unknown>
...
-          documentId: a.documentId,
-          requestBody: { requests: requests as any[] },
-        });
+        await withRetry(
+          (signal) => docs.documents.batchUpdate({
+            documentId: a.documentId,
+            requestBody: { requests: requests as any[] },
+          }, { signal }),
+          { ...ctx.runtimeConfig, retryMax: 0 },
+          'docs.documents.batchUpdate(findAndReplaceInDoc.multiline)',
```
piotr-agier: This hunk conflicts with master. #212 and #219 landed a revision lock around this exact call: `const lock = a.ifRevisionId ?? doc.data.revisionId`, `writeControlFor(lock)` in the request body, and a catch that maps a 400 on a locked write to the "revision mismatch, re-read and retry" error. The branch also predates #213 (`ifRevisionId` on `insertText` and `deleteRange`) and #220, so the current diff against master reads as reverting those.

Please rebase. The resolution is straightforward: keep master's `lock` and try/catch, and put the `withRetry(..., { ...ctx.runtimeConfig, retryMax: 0 }, ...)` call inside the `try` with `writeControlFor(lock)` still in the body. I ran the full gate on a trial rebase resolved that way and it is green (1035 tests, including the new write-control tests), so nothing else should move.

src/tools/docs.ts
```diff
@@ -4831,11 +5044,16 @@ export async function handleTool(toolName: string, args: Record<string, unknown>
...
+          fileId: a.documentId,
+          fields: 'id,name,description,mimeType,size,createdTime,modifiedTime,webViewLink,owners(displayName,emailAddress),lastModifyingUser(displayName,emailAddress),shared,parents,version',
+          supportsAllDrives: true,
+        }, { signal }),
+        ctx.runtimeConfig,
+        'drive.files.get(getDocumentInfo)',
+        ctx.log
+      );
 
       const file = response.data;
```
piotr-agier, line 5093: One call in this tool slipped through the sweep, presumably because it is chained across two lines: the revisionId lookup at lines 5070-5071 (`ctx.google.docs({...}).documents.get({ documentId: a.documentId, fields: 'revisionId' })`). It is a read inside a swallow-all try, so it is safe to give it the full read policy:

```ts
const docMeta = await withRetry(
  (signal) => ctx.google.docs({ version: 'v1', auth: ctx.authClient })
    .documents.get({ documentId: a.documentId, fields: 'revisionId' }, { signal }),
  ctx.runtimeConfig,
  'docs.documents.get(revisionId)',
  ctx.log
);
```

As written, a hung revision lookup still hangs `getDocumentInfo` with no deadline, which is the failure this PR exists to remove. With this fixed, the "every call site" claim in the title is true for the five files.

src/tools/docs.ts
```diff
@@ -4083,8 +4214,12 @@ export async function handleTool(toolName: string, args: Record<string, unknown>
...
           );
         }
-        const doc = await docs.documents.get({ documentId: a.documentId, includeTabsContent: true });
-        const text = collectDocPlainText(doc.data, a.tabId);
+        const doc = await withRetry(
+          (signal) => docs.documents.get({ documentId: a.documentId, includeTabsContent: true }, { signal }),
+          ctx.runtimeConfig,
+          'docs.documents.get(findAndReplaceInDoc.multiline)',
+          ctx.log
+        );        const text = collectDocPlainText(doc.data, a.tabId);
```
piotr-agier: The wrap's closing `);` and the next statement ended up on one line here and in two more places:

```
4222:  );        const text = collectDocPlainText(doc.data, a.tabId);
4270:  );        const text = collectDocPlainText(doc.data, a.tabId);
4318:  );          const hint = diagnoseZeroMatch(...);
```

Harmless at runtime and lint does not catch it, but it is a mechanical-edit artefact. Please split them onto their own lines.

src/tools/drive.ts
```diff
@@ -1551,15 +1688,20 @@ export async function handleTool(
...
+            requestBody,
+            media: {
+              mimeType: detectedMime,
+              body: mediaBody()
+            },
+            fields: 'id, name, size, mimeType, webViewLink',
+            supportsAllDrives: true
+          }, { signal }),
+          { ...ctx.runtimeConfig, retryMax: 0 },
+          'drive.files.create(upload)',
```
piotr-agier, line 1702: Worth calling out explicitly (applies equally to lines 1679, 977, 1034, 2124, 2155, 2316): for a non-stream response, the gaxios promise resolves only after the request body has been fully sent, so the 120s default now bounds the entire upload of `uploadFile`, `uploadPdfWithSplit`, `createTextFile`, `updateTextFile`, `createDocFromHTML`, and `restoreRevision`'s `files.update`. A large file on a slow link that used to succeed will now fail with `TimeoutError`, and the `createReadStream` body is left dangling.

I think the deadline is the right default, but "no behaviour change on a healthy call" is not quite true for these sites. Two options, either is fine by me:

1. Add a sentence to the CHANGELOG entry and PR body saying uploads are now bounded end to end and pointing at `--api-timeout=0`.
2. Give the upload sites a longer deadline, for example `{ ...ctx.runtimeConfig, retryMax: 0, apiTimeout: ctx.runtimeConfig.apiTimeout * N }`, if you would rather keep them tolerant of slow links.

CHANGELOG.md
```diff
@@ -13,6 +13,10 @@ All notable changes to this project will be documented in this file.
...
... because the policy counts `ECONNRESET` as retryable and a reset can be raised after a server has accepted a request, so a blind retry of a non-idempotent write could double-apply it — the same reasoning that already keeps HTTP 500 and 502 out of the retryable set. The deadline does real work rather than merely abandoning the wait: `withRetry` aborts the attempt through the `AbortSignal` it hands its callback, and every wrapped call forwards it as gaxios `{ signal }`. A metadata read a write tool makes to resolve a sheet id, a grid range, or a `textToFind` index is classified as the read it is. `createGoogleDoc`'s initial-content `batchUpdate`, the one pre-existing call, is left exactly as it was. Ten call sites in shared helper modules outside the five tool files (`driveImageUpload`, `text-content`, `download-file`, `auth/identity`) are not covered yet. No tool, parameter, or configuration option changes: the knobs already existed with their defaults (`--api-timeout`, 120s; `--retry-max`; `--retry-base-delay`), a healthy call behaves exactly as before, and `--api-timeout=0` restores the previous unbounded behaviour ([#218](https://github.com/piotr-agier/google-drive-mcp/pull/218))
```
piotr-agier: Two accuracy points for this entry and the PR body:

- A `TimeoutError` carries `code = 'ETIMEDOUT'`, which is in the retryable set, so a stalled *read* is retried `retryMax` times before the error surfaces. With defaults that is four attempts of 120s plus backoff, roughly eight minutes, not two. Still a big improvement over hanging forever, but "fails with a clear timed out after Nms" undersells the wait for reads. One sentence would do.
- The "not wrapped" list omits `src/index.ts`, which has six bare Drive calls in the context helpers (`resolveFolderId`, `checkFileExists`, and the search and read helpers at lines 384, 400, 441, 916, 940, 961, 977). Fine to leave for the follow-up, but please name them alongside the other four modules so the follow-up has a complete list.

src/tools/sheets.ts
```diff
@@ -885,10 +892,15 @@ async function batchUpdateOne(
...
-    spreadsheetId,
-    requestBody: { requests: [request] },
-  });
+  return withRetry(
+    (signal) => sheets.spreadsheets.batchUpdate({
+      spreadsheetId,
+      requestBody: { requests: [request] },
+    }, { signal }),
+    { ...ctx.runtimeConfig, retryMax: 0 },
+    'sheets.spreadsheets.batchUpdate',
```
piotr-agier: Nit. Every inline batchUpdate in this file carries an operation suffix such as `(addSheet)` or `(mergeCells)`, but this shared helper uses the bare label, and nine tools route through it (hide/showSheetDimension, the dimension-group tools, setColumnWidth, setRowHeight, autoResizeColumns/Rows). A timeout from any of them will read identically in the log.

`Object.keys(request)[0]` is right there and would give a free suffix:

```ts
`sheets.spreadsheets.batchUpdate(${Object.keys(request)[0]})`,
```

piotr-agier (commented):
Verified the mechanical part end to end and it is correct. All 176 sites: the read/write split matches the table exactly (86 full retries, 90 with `retryMax: 0`), every closure holds a single API call, `{ signal }` is in the gaxios options slot everywhere, and the only non-wrap changes are threading `ctx` into `resolveGridRange` and `reconcileCreatedPermissionRole`. The cancellation claim also holds: googleapis-common merges per-call options into the gaxios request, and the gaxios copy it bundles combines `opts.signal` with its own timeout signal.

The `restoreRevision` stream reasoning checks out too: the promise settles on headers, the timer is cleared on success, and the controller is never aborted after a successful attempt, so the body transfer is untouched. The test file is well built, and the env-var timing works because the server module is imported dynamically inside `setupTestServer`.

Approving once the rebase lands and the two docs.ts fixes (the unwrapped revisionId read, the joined lines) are in. The upload-deadline and changelog points are should-fix, not blockers.

salimaCodebase:
Rebased onto current master (c61b1a6, release 2.11.0). New head: 6c732b6. The only conflict was in `src/tools/docs.ts`, in the multi-line `findAndReplaceInDoc` batch write: kept master's revision-lock request body (`writeControlFor(lock)`) and try/catch mapping a locked-write 400 to a revision-mismatch error, and re-applied only the `withRetry(..., { ...ctx.runtimeConfig, retryMax: 0 }, ...)` wrapper around the `batchUpdate` call. Full CI gate (lint, build, registry:validate, test) is green: 1035/1035 tests pass, including the documentation-reference test and the #220 revisionId-in-fields-mask gate.

src/tools/drive.ts
salimaCodebase: files.create and files.update carry media: { body: data.content }, and for a non-stream response the gaxios promise settles only after the request body has been fully sent, so the 120s default now bounds the whole upload transfer, not just time-to-response. This deployment's upload endpoint accepts up to 1 GiB, and no ordinary link moves 1 GiB in 120s, so a large upload that used to succeed now fails outright with TimeoutError. Three ways to resolve this: (1) leave media-carrying calls unwrapped, so uploads keep today's unbounded behaviour and the PR says so explicitly; (2) give upload sites a separate, larger deadline (e.g. { ...ctx.runtimeConfig, retryMax: 0, apiTimeout: ctx.runtimeConfig.apiTimeout * N }); or (3) keep the current 120s bound and document it as a breaking change for large uploads. I'd lean toward (2), since it keeps hang protection on every upload site without silently failing a large one that used to work, but leaving this to your call with the concrete number (1 GiB / 120s) in front of you rather than picking for you.

salimaCodebase: The three gating items (rebase onto master, the getDocumentInfo revisionId wrap, the joined lines) landed in 0feb53d, before this reply — that commit is on the branch now at 0d480e8.

For the upload deadline: went with option 2. Added UPLOAD_TIMEOUT_MS = 30 minutes in src/utils/retry.ts (a plain constant, not a new config knob) and pointed every media-carrying call in the five tool files at it (createTextFile, updateTextFile, uploadFile, uploadPdfWithSplit, restoreRevision's rewrite, createDocFromHTML — eight call sites, not just the files.create/files.update pair) via apiTimeout: UPLOAD_TIMEOUT_MS instead of the 120s default; retryMax stays 0. 30 minutes covers a 1 GiB upload on an ordinary link while still bounding a true stall. Switching this to unbounded (apiTimeout: 0) or to a config knob is a one-line change per call site if you'd rather have either.

salimaCodebase:
Rebased onto current master (29b1172). New head: 70c4184.

The only conflict was master's #210 (getGoogleSheetCells + src/tools/sheetCells.ts), unseen by this branch. Two resolutions:

- src/tools/sheets.ts: import-block conflict (this branch's `withRetry` import next to #210's new `sheetCells.js` imports) — kept both.
- getGoogleSheetCells's `spreadsheets.get` was unwrapped on master; wrapped it under the read policy (full `withRetry`, since it's a plain read) as `sheets.spreadsheets.get(cells)`, matching every other Sheets read in the file.
- CHANGELOG.md: kept #210's entry under Added and this PR's under Changed, in Unreleased; corrected the call-site count while there (179 total, 89 reads, 90 writes — one higher than previously stated, since `getGoogleSheetCells` is now included).

sheetCells.ts itself is pure helper code (address arithmetic, field-mask building, matching, truncation) with no Google API calls, so nothing to wrap there.

Also fixed test/tools/sheet-cells-handler.test.ts's fakeCtx, which had no `runtimeConfig`/`log` and failed once the call it stubs went through `withRetry`.

Full gate green: lint, build, registry:validate, and 1086/1086 tests.

